PR #106, opened 2017-02-06
Gcw 1410 integration test suite

Darshan-Chauhan:


app/components/partial-designate-input.js
```diff
@@ -1,24 +1,28 @@
...
-      var setItems = this.get('item.setItem.items');
-      var minQty = setItems.canonicalState[0]._data.quantity;
-      setItems.canonicalState.forEach(record =>{
-        var qty = record._data.quantity;
-        if(qty < minQty) {
-          minQty = qty;
-        }
-      });
-      return minQty;
+    if(this.get("ENV.environment") !== "test") {
```
steveyken: @Darshan-Chauhan is there a way to remove this constraint - if we do the tests differently?

tests/integration/components/auto-resize-textarea-test.js
```diff
@@ -0,0 +1,36 @@
...
+test('is of text type', function(assert) {
+  assert.expect(1);
+  assert.equal($('#description')[0].type, "textarea");
+});
+
+test('initial height is 70px', function(assert) {
+  assert.expect(1);
+  assert.equal($('#description').css('height'), "70px");
+});
+
```
steveyken: Great to see an example of a component test!

app/components/partial-designate-input.js
Darshan-Chauhan: @steveyken I'll check if there is any work around for this.

PR #112, opened 2017-02-17
Gcw 1433 integration test suite for partial move

Darshan-Chauhan:


tests/acceptance/undispatch-select-location-test.js
```diff
@@ -0,0 +1,67 @@
...
+    item1 = FactoryGuy.make("item", { state: "submitted", quantity: 10 , designation: designation});
+    orders_package1 = FactoryGuy.make("orders_package", { state: "dispatched", quantity: 4, item: item1, designation: designation });
+
+    visit("/orders/");
+
+    fillIn('#searchText', designation.get("code"));
+    andThen(function(){
+      click(find('ul.list li:first div.row.state_details'));
+    });
+
```
steveyken: should we be checking some output here?

tests/acceptance/partial-move-test.js
```diff
@@ -0,0 +1,121 @@
...
+    App = startApp({}, 2);
+    TestHelper.setup();
+
+    designation = FactoryGuy.make("designation", { state: "closed" });
+    item = FactoryGuy.make("item", { state: "submitted", quantity: 5 });
+    package_location = FactoryGuy.make("packages_location", { quantity: 5 , item: item});
+    orders_package = FactoryGuy.make("orders_package", { state: "designated", quantity: 5, item: item, designation: designation });
+
+    mockFindAll('designation').returns({json: {designations: [designation.toJSON({includeId: true})]}});
+    visit("/items/" + item.id + "/partial_move");
```
steveyken: we're checking we can get to the partial move screen here - do you want to check contents of the page too?

tests/acceptance/partial-move-test.js
```diff
@@ -0,0 +1,121 @@
...
+        }
+    });
+  },
+  afterEach: function() {
+    Ember.run(function() { TestHelper.teardown(); });
+    Ember.run(App, 'destroy');
+  }
+});
+
+test("Partially moving a Package", function(assert) {
```
steveyken: :+1: this is a thorough test :)

tests/acceptance/undispatch-select-location-test.js
```diff
@@ -0,0 +1,67 @@
...
+    assert.equal(currentPath(), "items.search_location");
+    //selecting first location
+    click($('ul.list-activity li:first'));
+    andThen(function() {
+      //clicking on move of messageBox
+      click($('div#messageBox:eq(5) a:last'));
+    });
+  });
+  andThen(function() {
+    assert.equal(currentPath(), "items.detail");
```
steveyken: do you want to check the object has been updated too? i.e. item status should now be "in stock"

tests/integration/components/partial-move-input-test.js
```diff
@@ -0,0 +1,35 @@
+import { test, moduleForComponent } from 'ember-qunit';
```
steveyken: so my comments above were about checking the partial_move screen but now I see you are testing the details component here instead - a much better idea!

Can you inject any mock data into the partial move component test so we can check for line items i.e. move qty X  from package Y in location Z and move qty A from package B in location C

tests/acceptance/partial-move-test.js
Darshan-Chauhan: No @steveyken , as it is beforeEach block(although we can write some checks here but it will be unconventional). I've checked the page contents in test cases.

tests/acceptance/undispatch-select-location-test.js
Darshan-Chauhan: ^

tests/acceptance/undispatch-select-location-test.js
Darshan-Chauhan: Yes, that would be better. I'll add that.

tests/integration/components/partial-move-input-test.js
Darshan-Chauhan: @steveyken , we can add mock data but checking for line items won't be possible, as you can see to even check the component I'd to render that component in beforeEach block.
- If we want to check for line-items then we would have to render other buttons message-box and may be entire page.
- So in integration testing, I think we can only write test cases which are related to component but we can't preform any other operations for which we need to render other details of page
- As for moving quantity for packages in different locations, I've already covered that in partial-move-test.js file
- Also we're dealing with singleton packages only, so it is not possible to move partial quantity so there would be always **one line item only**, so that's why scenarios for test-cases are few.

Let me know if any scenarios are still missing or there are other concerns :)
Thanks

mattgow: Steve says this is fine, thanks @Darshan-Chauhan

PR #124, opened 2017-03-30
GCW-1505-mask-qty-items

Darshan-Chauhan:


app/templates/orders/orders_packages.hbs
```diff
@@ -40,14 +40,23 @@
         </div>
       </div>
     {{else}}
-      {{#link-to 'items.detail' ordersPackage.item.id }}
+      {{#if ordersPackage.item.isSingletonItem}}
```
steveyken: shame to have to duplicate the nested template just to accomodate the if/else clause. Would the 'disabled' attribute of link-to (https://emberjs.com/api/classes/Ember.Templates.helpers.html#method_link-to) work here or does that disabled everything inside the component? What about setting the route to '' if quantity package or creating a small partial template to reduce duplication?

Darshan-Chauhan: @steveyken , thanks for pointing that out. It definitely reduces the code. `disabled` attribute of link-to just works fine so need to set different route or partial template. I've updated the code. You can check it and let me know if any more changes are required. 
Thanks

PR #125, opened 2017-03-31
Gcw 1466 refine packages location move ui

namrataukirde:


app/components/partial-move-input.js
```diff
@@ -31,15 +38,59 @@ export default Ember.TextField.extend({
     }
     var regex = /^\d+$/;
     var input_value = this.get('value');
+    var siblingPackagesLocations = this.get('item.siblingPackagesLocations');
+    var isInvalid;
```
Darshan-Chauhan: Do we need to assign any default value?

app/controllers/items/partial_move.js
```diff
@@ -32,6 +35,19 @@ export default Ember.Controller.extend({
...
+    },
+    editPackagesLocation(packagesLocation){
+      this.set('isEditing', true);
+      packagesLocation.set('isEditing', true);
+    },
+
+    notNow(item){
+      var editedPackagesLocations = item.get('packagesLocations').filterBy('isEditing', true);
+      editedPackagesLocations.forEach(packagesLocation => {
+        Ember.$(`#packages-qty-location-${packagesLocation.id}`)[0].innerText = packagesLocation.get('quantity');
```
Darshan-Chauhan: Can we use text( ) function and pass _packagesLocation.get('quantity')_ as innerText is pure JS?

PR #133, opened 2017-04-21
GCW-1433 test cases for partial move for singleton package

Darshan-Chauhan:


tests/acceptance/partial-move-test.js
```diff
@@ -28,94 +29,95 @@ module('Acceptance: Partial move', {
...
+    mockFindAll('item').returns({ json: {items: [item.toJSON({includeId: true})], orders_packages: [orders_package.toJSON({includeId: true})], meta: {search: item.get('inventoryNumber').toString()}}});
+
+    $.mockjax({url: '/api/v1/stockit_item*', type: 'GET', status: 200,responseText: {
+          items: [item.toJSON({includeId: true})]
+        }
+    });
+
+    //visiting Item's partial move
+    visit("/items/" + item.id + "/partial_move");
+
```
steveyken: so we want to check there is expected output on the page here? We are going to the partial_move page but not checking for anything other than clicking the move link

Darshan-Chauhan: @steveyken I've to go to partial_move URL for every test case.. So I thought it would be a good Idea to put mock urls and link to partial_move in beforeEach block..
We're not checking for anything in beforeEach block, it just gets executed before every test case

PR #140, opened 2017-05-26
Gcw 1553 skip move screen for singleton items

Darshan-Chauhan:


app/controllers/items/search_location.js
```diff
@@ -105,16 +107,33 @@ export default searchModule.extend({
...
-      var totalQty = localStorage["totalQty"];
+      var packagesLocationQty = [];
+      var totalQty = '';
+      if(this.get('skipScreenForSingletonItem')) {
+        var record = {};
+        var packages_location = this.get('store').peekRecord('packages_location', this.get('pkgsLocationId'));
+        record["packages_location_id"] = this.get('pkgsLocationId');
+        record["location_id"] = packages_location.get('locationId');
+        record["package_id"] = packages_location.get('packageId');
+        record["new_qty"] = "1";
```
namrataukirde: Instead of defaulting it to 1 can we take quantity of package? I know we are handling only single quantity items from stock app. But still I feel if we are moving item then we should send package quantity instead of assigning it 1.  Just suggestion.

Darshan-Chauhan: Is package is designated then package.qty will be 0 so that would be a problem so defaulting 1 is better

Darshan-Chauhan: if we assign **received_quantity** then it looks like we're moving received_quantity which doesn't read right, so I think defaulting it to 1 is still our best option

namrataukirde: I dont think it looks right. But if you think it's fine then keep it.

PR #141, opened 2017-05-31
GCW-1514-added-query-param-to-show-quantity-item

namrataukirde:


app/controllers/items/index.js
```diff
@@ -4,9 +4,10 @@ import searchModule from "../search_module";
 
 export default searchModule.extend({
 
-  queryParams: ['searchInput', "itemSetId"],
+  queryParams: ['searchInput', "itemSetId", "showQuantityItems"],
```
steveyken: This works well for the single case. How would we make this work at the generalised level? Say if there were cases where we wanted to add a global flag? Wouldn't want to have to add this into every controller that needed it. Is there a general way to add a whitelisted query string argument to the `Ember.Controller` base class?

PR #143, opened 2017-06-09
Gcw 1514 show qty items

Darshan-Chauhan:


Darshan-Chauhan:
-  They are failing because they were written for qty Items. We've disabled all the actions(dispatch designate etc) for such Items and also **showQuantityItems** and disabled actions are not related. 
- Even we if made changes such that we include showQuantityItems feature, it still won't work.
- Because those disabled actions are not related to this feature. means when you enter **showQuantityItems** in url, it doesn't enable/disable those actions. Also enabling the actions might cause an issue because we've made a lot of changes regarding to singleton packages

PR #173, opened 2017-08-21
GCW-1631 send numbers for Donor Condition

Darshan-Chauhan:


app/components/inline-select-list.js
```diff
@@ -17,6 +17,16 @@ export default SelectList.extend({
     return { id: `${itemCondition.charAt(0)}` };
   }),
 
+  changeDonorCondition(conditionId) {
```
namrataukirde: Instead of doing this way can we add directly numbers as value of select option ?

PR #197, opened 2017-10-23
show msg on quantity validation failure

namrataukirde:


app/components/designate-form.js
```diff
@@ -236,8 +236,13 @@ export default Ember.Component.extend({
             loadingView.destroy();
             this.get('router').transitionTo("items.index");
           }
-        })
-        .finally(() => {
+        }).catch((error) => {
+          if(error.status === 422){
+            var errors = Ember.$.parseJSON(error.responseText).errors;
+            this.get("messageBox").alert(errors);
+            this.get('router').transitionTo("items.index");
```
Darshan-Chauhan: Can we add this to callback of mesageBox?

namrataukirde: Yes Sure 👍

PR #198, opened 2017-10-27
GCW 1577 show organisation details in order

viveksriram1:


app/templates/orders/detail.hbs
```diff
@@ -51,7 +51,11 @@
           Organisation:
         </div>
         <div class="small-8 columns">
-          {{apply-line-break model.organisation.name}}
```
Darshan-Chauhan: I think we can use inline **if** condition here.
What do you think @viveksriram1 ?

app/models/organisation.js
```diff
@@ -3,6 +3,7 @@ import attr from 'ember-data/attr';
 
 export default Model.extend({
```
Darshan-Chauhan: Can we remove this extra line??

app/models/organisation.js
```diff
@@ -3,6 +3,7 @@ import attr from 'ember-data/attr';
...
 
-  name: attr('string')
+  name: attr('string'),
+  nameEn: attr('string')
```
Darshan-Chauhan: ^

app/templates/orders/detail.hbs
viveksriram1: I tried inline if condition but its not showing properly because of apply-line-break

app/models/organisation.js
viveksriram1: sure

PR #210, opened 2017-11-27
Revert "fix minifield url for staging for rollbar"

namrataukirde:
Reverts crossroads/stock.goodcity#209


namrataukirde:
This is not needed. thus closing it.

PR #229, opened 2018-02-12
GCW 1367 build purposes for order

Darshan-Chauhan:
Hi Team,

This PR contains changes for ticket https://jira.crossroads.org.hk/browse/GCW-1367.
- Added new OrdersPurpose and Purpose model
- Added new order's purpose view
- Added unit and acceptance test cases for the same


app/helpers/is-last-word-in-array.js
```diff
@@ -0,0 +1,5 @@
+import Ember from "ember";
+
+export default Ember.Helper.helper(function(leftside) {
+  return (leftside[1] + 1 < leftside[0]) ? true : false;
```
shivanibhanwal: can we write it this way? Do you see any issue? if not that having it in this manner would be good as we are having an expression here
`(leftside[0] > leftside[1] + 1) ? false : true;`

app/models/designation.js
```diff
@@ -21,7 +22,7 @@ export default Model.extend({
   items:              hasMany('item', { async: true }),
   ordersPackages:     hasMany('ordersPackages', { async: false }),
   orderTransport:     belongsTo('orderTransport', { async: false }),
-  // purposes:           hasMany('purpose', { async: false }),
+  ordersPurposes:     hasMany('orders_purpose', { async: false }),
```
shivanibhanwal: Why the Name has inconsistancy?? 
e.g. `belongsTo('orderTransport'` but now we have ` ordersPurposes:     hasMany('orders_purpose'` 
Ideally in JS it should use camelCase `ordersPurpose`
is not it?

Darshan-Chauhan: Yes @shivanibhanwal , Thanks for pointing it out. I'll change it.

PR #269, opened 2018-07-04
GCW 1998 xcode upgrade

Darshan-Chauhan:
Hi Team,
This PR upgrades the following.
- Cordova-ios to `4.5.4`
- XCode to `9.2`
- Cordova to lastest version

Downgrades:
- Ember-cli to 2.11(as it was failing with node engine needed >=4.5 || 6 error, which we already have but 2.11 seemed to be working so downgraded to 2.11)

Removes from Circle.yml
- Cordova(as we're installing cardova in rakefile so not needed)
- ember-cli(installing through package.json so not needed)
- Watchmen(only needed to run test cases, in case watchmen is not installed them ember test falls back to node which is fine)

Changes needed to make XCode `9.2` get working on CI
- Add `developmentTeam` id to build.json or directly to cordova build command (developmentTeam id can be found here => https://developer.apple.com/account/#/membership/6B8FS8W94M )
- From XCode 8 and onwards we need to create `iPhone Development` cert instead of `iPhone Distribution` 
- I created `IOS Development` certificate which can be found here => https://developer.apple.com/account/ios/certificate/development (date: Jul 04, 2019)
- .p12 info can be found on ticket https://jira.crossroads.org.hk/browse/GCW-1998
- I created provisioning profile using this development certificate. To download provisioning profile => https://developer.apple.com/account/ios/profile/limited
- I later uploaded certificate and provisioning profile to CI and specified `developmentTeam` id `provisioningProfile` accordingly.

Reference for Cordova build options:- https://github.com/apache/cordova-ios/blob/master/bin/templates/scripts/cordova/build#L34

**Note**:- I was unable to use build.json file(in cordova root dir) in the build so I'd to add changes directly to Rakefile, if someone finds solution for this then please create a PR for the same.

Please review this PR and let me know if any other changes are require.
Thanks :)

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -130,9 +130,9 @@ namespace :cordova do
   desc "Cordova build {platform}"
   task build: :prepare do
     Dir.chdir(CORDOVA_PATH) do
-      team_id = '6B8FS8W94M'
-      provisioning_profile = '4e899d06-4e25-40b5-8ea3-9d93822e8c55'
-      provisioning_profile = '866b46e8-096e-46a0-97fe-0579091c887e' if(environment === "production")
+      team_id = ENV['IOS_DEVELOPMENT_TEAM_ID']
+      provisioning_profile = ENV['PROVISIONING_PROFILE_STAGING']
+      provisioning_profile = ENV['PROVISIONING_PROFILE_PROD'] if(environment === "production")
       build = (environment == "staging" && platform == 'android') ? "debug" : "release"
       system({"ENVIRONMENT" => environment}, "cordova compile #{platform} --#{build} --device --codeSignIdentity='iPhone Developer' --developmentTeam=#{team_id} --provisioningProfile=#{provisioning_profile}")
     end
--- a/Rakefile
+++ b/Rakefile
@@ -103,7 +103,7 @@ end
 namespace :cordova do
   desc "Install cordova package globally"
   task :install do
-    sh %{ npm list --depth 1 --global cordova; if [ $? -ne 0 ]; then npm install -g cordova@6.5.0; fi }
+    sh %{ npm list --depth 1 --global cordova; if [ $? -ne 0 ]; then npm install -g cordova; fi }
     sh %{ npm list --depth 1 --global cordova-update-config; if [ $? -ne 0 ]; then npm install -g cordova-update-config; fi }
   end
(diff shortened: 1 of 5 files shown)


shivanibhanwal:
I really love this detailed PR comment. @Darshan-Chauhan. Very informative and useful for everyone 👍

Rakefile
```diff
@@ -130,8 +130,11 @@ namespace :cordova do
   desc "Cordova build {platform}"
   task build: :prepare do
     Dir.chdir(CORDOVA_PATH) do
+      team_id = ENV['IOS_DEVELOPMENT_TEAM_ID']
+      provisioning_profile = ENV['PROVISIONING_PROFILE_STAGING']
+      provisioning_profile = ENV['PROVISIONING_PROFILE_PROD'] if(environment === "production")
       build = (environment == "staging" && platform == 'android') ? "debug" : "release"
-      system({"ENVIRONMENT" => environment}, "cordova compile #{platform} --#{build} --device")
+      system({"ENVIRONMENT" => environment}, "cordova compile #{platform} --#{build} --device --codeSignIdentity='iPhone Developer' --developmentTeam=#{team_id} --provisioningProfile=#{provisioning_profile}")
```
steveyken: Given we have the `codeSignIdentity`, `developmentTeam` and `provisionProfile` details in `cordova/build.json` do we need them here as well or can we just use `build.json` settings based on the environment selected - `debug` or `release`?

Darshan-Chauhan:
@shivanibhanwal , glad to know that it was helpful. :)

Rakefile
Darshan-Chauhan: @steveyken , yes using build.json would be ideal, and we can specify config file using this https://github.com/apache/cordova-ios/blob/master/bin/templates/scripts/cordova/build#L48 , 
But I tried it and got the error saying build.json file doesn't exist. 

Also tried creating one build.json and specifying the same configuration but still got the same error.

Also I noticed that build.json was using old provisioningProfile(we did not change it after last time we updated our pro. profiles), so if build.json was getting used during cordova build then none our build would've worked. So I suspect that cordova wasn't picking up configuration from build.json from the beginning.

And as upgrading to XCode was more important so I thought I should do that first and later we can figure the way to use build.json

Thanks

PR #284, opened 2018-08-24
Code climate test coverage changes

Darshan-Chauhan:
Hi Team,

This PR uses CodeClimate's **test-reporter** as it has support for parallel builds, multiple languages, and will receive ongoing support from CodeClimate in the future. 
The reason for change is we were using  older, deprecated test reporter which wasn't showing code-coverage correctly so we're hoping to fix the issue with this.

Test-reporter:- https://github.com/codeclimate/test-reporter
Configuration:- https://docs.codeclimate.com/docs/configuring-test-coverage

Please take a look and let me know if any further changes are required.
Thanks


Darshan-Chauhan:
@steveyken , I was just testing with apexcov. I don't think we need that as we can directly upload to CodeClimate. I'll remove it.

PR #289, opened 2018-09-11
GCW 2118 new order search layout

patrixr:
New layout for order listing

Notes: 
- StockIt orders' transport types are currently shown as "Other"
- Built using Fontawesome v4, still requires an update to v5
- Infinite scroll is still not working as intended (as discussed in a separate thread)

### DIFF
--- a/app/controllers/orders/index.js
+++ b/app/controllers/orders/index.js
@@ -4,5 +4,16 @@ export default searchModule.extend({
 
   searchModelName: "designation",
   unloadAll: true,
-  minSearchTextLength: 2
+  minSearchTextLength: 2,
+
+  onItemLoaded: function (record) {
+    if (record.get("gcOrganisationId")) {
+      this.store.findRecord("gc_organisation", record.get("gcOrganisationId"));
+    }
+    if(record.constructor.toString() === "stock@model:designation:") {
+      this.store.query("orders_package", { search_by_order_id: record.get("id") });
+    }
+  }
+
+
 });
--- a/app/controllers/orders/index.js
+++ b/app/controllers/orders/index.js
@@ -7,8 +7,9 @@ export default searchModule.extend({
   minSearchTextLength: 2,
 
   onItemLoaded: function (record) {
-    if (record.get("gcOrganisationId")) {
-      this.store.findRecord("gc_organisation", record.get("gcOrganisationId"));
+    const orgId = record.get("gcOrganisationId");
+    if (orgId) {
+      this.store.findRecord("gc_organisation", orgId, { reload: false });
     }
     if(record.constructor.toString() === "stock@model:designation:") {
       this.store.query("orders_package", { search_by_order_id: record.get("id") });
--- a/app/models/designation.js
+++ b/app/models/designation.js
@@ -4,6 +4,7 @@ import attr from 'ember-data/attr';
 import { belongsTo, hasMany } from 'ember-data/relationships';
 
 export default Model.extend({
+  i18n: Ember.inject.service(),
   utilityMethods: Ember.inject.service(),
 
   status:               attr('string'),
@@ -60,6 +61,59 @@ export default Model.extend({
   capitalizedState: Ember.computed('state', function() {
     return this.get("state").capitalize();
   }),
+
+  stateIcon: Ember.computed('stateIcon', function () {
+    const state = this.get("state");
+    switch (state) {
+      case "awaiting_dispatch":
+      case "scheduled":
+        return "clock-o";
+      case "processing":
+        return "list";
+      case "submitted":
+        return "envelope";
+      case "dispatching":
+        return "paper-plane";
+      case "cancelled":
+        return "thumbs-down";
+      case "closed":
+        return "lock";
+      default:
+        return "";
+    }
+  }),
+
+  transportIcon: Ember.computed("transportIcon", function() {
+    const key = this.get("transportKey");
+    switch (key) {
+      case "gogovan_transport":
+        return "truck";
+      case "collection_transport":
+        return "male";
+      default:
+        return "";
+    }
+  }),
+
+  transportLabel: Ember.computed("transportLabel", function() {
+    console.log(this.get("contact"));
+    const key = this.get('transportKey');
+    return this.get("i18n").t(`order_transports.${key}`);
+  }),
+
+  transportKey: Ember.computed("transportKey", function() {
+    const transportType = this.get('orderTransport.transportType');
+    if (transportType) {
+      // GoodCity Order
+      if (transportType === "ggv") {
+        return "gogovan_transport";
+      } else if (transportType === "self") {
+        return "collection_transport"
+      }
+    }
+    // TODO: StockIt orders
+    return "unknown_transport"
+   }),
 
   ordersPackagesCount: Ember.computed('ordersPackages.[]', 'ordersPackages.@each.quantity', 'ordersPackages.@each.state', function() {
     return this.get("ordersPackages").filterBy('quantity').length;
--- a/app/models/designation.js
+++ b/app/models/designation.js
@@ -96,7 +96,6 @@ export default Model.extend({
   }),
 
   transportLabel: Ember.computed("transportLabel", function() {
-    console.log(this.get("contact"));
     const key = this.get('transportKey');
     return this.get("i18n").t(`order_transports.${key}`);
   }),
--- a/app/styles/templates/orders/_index.scss
+++ b/app/styles/templates/orders/_index.scss
@@ -1,5 +1,106 @@
+@import "../../_colors";
+
+$header-height: 3em;
+
 .orders_search_result {
   padding-top: 3rem;
+  
+  li.order_block {
+    border: none;
+    padding-left: 0;
+    padding-right: 0;
+
+
+    .order_process_summary .value {
+      >span {
+        margin-right: 0.7em;
+      }
+    }
+
+
+    // ------------
+    // HEADER
+    // ------------
+
+    .order_header {
+      background-color: #384F72;
+      padding-left: 1em;
+      padding-top: 0;
+      height: $header-height;
+      line-height: $header-height;
+
+      .order_transport, .order_code {
+        color: white;
+
+        .fa {
+          margin-right: 0.5em;
+          margin-left: 0.5em;
+        }
+      }
+
+      .order_state {
+        text-transform: capitalize;
+
+        >div {
+          float: right;
+        }
+
+        .order_state_text {
+          padding-right: 1em;
+        }
+
+        .order_state_icon {
+          color: white;
+          height: $header-height;
+          width: $header-height;
+
+          .fa {
+            margin-left: 1em;
+          }
+
+          &.submitted {
+            background-color: $submitted-state-color;
+          }
+          &.scheduled, &.awaiting_dispatch {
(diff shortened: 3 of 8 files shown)


app/controllers/orders/index.js
```diff
@@ -4,5 +4,17 @@ export default searchModule.extend({
 
   searchModelName: "designation",
   unloadAll: true,
-  minSearchTextLength: 2
+  minSearchTextLength: 2,
+
+  onItemLoaded: function (record) {
```
Darshan-Chauhan: @patrixr , with ES6 syntax you can directly declare function like this
```
onItemLoaded(record) {
  ---  
}
```

app/models/designation.js
```diff
@@ -61,6 +62,58 @@ export default Model.extend({
     return this.get("state").capitalize();
   }),
 
+  stateIcon: Ember.computed('stateIcon', function () {
```
Darshan-Chauhan: @patrixr , I think this computed property should be dependent on `order.state` ?
What do you think?

app/models/designation.js
```diff
@@ -61,6 +62,58 @@ export default Model.extend({
...
+      case "cancelled":
+        return "thumbs-down";
+      case "closed":
+        return "lock";
+      default:
+        return "";
+    }
+  }),
+
+  transportIcon: Ember.computed("transportIcon", function() {
```
Darshan-Chauhan: @patrixr , I think this computed property should be dependent upon `transportKey `

app/models/designation.js
```diff
@@ -61,6 +62,58 @@ export default Model.extend({
...
+      case "gogovan_transport":
+        return "truck";
+      case "collection_transport":
+        return "male";
+      default:
+        return "";
+    }
+  }),
+
+  transportLabel: Ember.computed("transportLabel", function() {
```
Darshan-Chauhan: @patrixr , same as above

app/models/designation.js
```diff
@@ -61,6 +62,58 @@ export default Model.extend({
...
+    const key = this.get('transportKey');
+    return this.get("i18n").t(`order_transports.${key}`);
+  }),
+
+  transportKey: Ember.computed("transportKey", function() {
```
Darshan-Chauhan: @patrixr , you can make this computed property dependent like this
`transportKey: Ember.computed("orderTransport.transportType", function() {`

app/models/designation.js
```diff
@@ -61,6 +62,58 @@ export default Model.extend({
...
+    const transportType = this.get('orderTransport.transportType');
+    if (transportType) {
```
Darshan-Chauhan: I don't think we need this if condition here.

app/styles/templates/orders/_index.scss
```diff
@@ -1,5 +1,106 @@
+@import "../../_colors";
```
Darshan-Chauhan: This file should be imported in `app.scss` file so it'll be available for other pages. We'll need to use these colors on multiple pages.

app/templates/orders/order_block.hbs
```diff
@@ -1,31 +1,82 @@
...
-      <div class="small-5 columns">
-        {{order.code}} - {{order.status}}
-      </div>
-      <div class="small-4 columns">
-        {{order.designatedOrdersPackages.length}} <i class="fa fa-shopping-basket"></i>
-        {{order.dispatchedOrdersPackages.length}} <i class="fa fa-ship"></i>
+    <div class="row order_header">
+
+      <div class="small-4 columns order_transport"> 
+        <i class="fa fa-lg fa-{{ order.transportIcon }}"></i> {{ order.transportLabel }}
```
Darshan-Chauhan: I think we can remove spaces from curly braces

app/templates/orders/order_block.hbs
```diff
@@ -1,31 +1,82 @@
...
-    <div class="row">
-      <div class="small-12 columns">
-        {{order.stockitContact.fullName}}: {{order.stockitContact.displayNumber}}
-        {{#if order.localOrder.clientName}}
-          for {{order.localOrder.clientName}}
-        {{/if}}
-      </div>
+    <div class="row order_body">
+
+      {{#if order.orderTransport.scheduledAt }}
```
Darshan-Chauhan: @patrixr , correct me if I'm wrong. For StockIt Orders, OrderTransport doesn't exist right? So it'll fail for StockIt orders. So I think we can add a check for that here. Unfortunately Handlebars doesn't support `and | or` or any other conditions with if so we've to write helpers for such cases. I think you can use the below line

`{{#if (is-and order.orderTransport order.orderTransport.scheduledAt)}}`

Here we've used `is-and` helper which you can find in **goodcity-lib** repo.

app/templates/orders/order_block.hbs
```diff
@@ -1,31 +1,82 @@
...
+          <div class="small-4 columns field">
+            {{ t "order.submitted_by" }}
+          </div>
+          <div class="small-8 columns value">
+            {{ order.submittedBy.firstName }} {{ order.submittedBy.lastName }}
+          </div>
+        </div>
+      {{/if}}
+      
+      {{#if order.gcOrganisationId or order.organisationId }}
```
Darshan-Chauhan: Can we use
`{{#if (is-or order.gcOrganisationId or order.organisationId)}}`

app/templates/orders/order_block.hbs
```diff
@@ -1,31 +1,82 @@
...
+        <div class="row">
+          <div class="small-4 columns field">
+            {{ t "organisation_title" }}
+          </div>
+          <div class="small-8 columns value">
+            {{ if order.isGoodCityOrder order.gcOrganisation.nameEn order.organisation.name }}
```
Darshan-Chauhan: @patrixr , I'm having trouble understanding this condition. Can you please explain?

app/templates/orders/order_block.hbs
```diff
@@ -1,31 +1,82 @@
...
+          <div class="small-4 columns field">
+            {{ t "order.submitted_for_client" }}
+          </div>
+          <div class="small-8 columns value">
+            {{order.stockitContact.fullName}}
+          </div>
+        </div>
+      {{/if}}
+
+      {{#if order.designatedOrdersPackages.length or order.dispatchedOrdersPackages.length }}
```
Darshan-Chauhan: You can use helper here
`{{#if (is-or order.designatedOrdersPackages.length order.dispatchedOrdersPackages.length)}}`

app/models/designation.js
patrixr: @Darshan-Chauhan I see, I mis-used the ember computed properties. I'll fix those :)

app/templates/orders/order_block.hbs
patrixr: @Darshan-Chauhan I was under the impression handlebars handled undefined variables. I've tested it here http://jsfiddle.net/x90Lbe38/2/

I'll switch to is-and to be safe, but it'd be good to check it out :)

app/templates/orders/order_block.hbs
patrixr: @Darshan-Chauhan this condition extracts the name from the gc_organisation only if it is a good city order, otherwise uses the stockit_organisation name

app/models/designation.js
patrixr: It was just for structure, as the else block would theoretically handle transports for stockit orders. I can remove it, and re-add it once(if) we move to stockit orders

app/templates/orders/order_block.hbs
patrixr: Just for readability

app/templates/orders/order_block.hbs
patrixr: Will do

app/controllers/orders/index.js
```diff
@@ -4,5 +4,17 @@ export default searchModule.extend({
...
   unloadAll: true,
-  minSearchTextLength: 2
+  minSearchTextLength: 2,
+
+  onItemLoaded(record) {
+    const orgId = record.get("gcOrganisationId");
+    if (orgId) {
+      this.store.findRecord("gc_organisation", orgId, { reload: false });
+    }
+    if(record.constructor.toString() === "stock@model:designation:") {
```
patrixr: @Darshan-Chauhan Would you know what this condition is about ? (was previously in search_module.js)

Darshan-Chauhan: Yeah, I think when searching orders, we were having trouble loading OrdersPackages associated with Order. So if returned record is designation model then we're requesting ordersPackages associated with that designation object.

But I think we're getting ordersPackages data with Orders now so we don't need to fire extra request for ordersPackages.

@patrixr , you can check if we're getting OrdersPackages in Order search response then we can remove this.

namrataukirde:
Hey @patrixr PR looks good. :)

Are you planning to add test cases?

patrixr:
> Hey @patrixr PR looks good. :)
> 
> Are you planning to add test cases?

@namrataukirde I was thinking of testing the model's computed properties, the rest is mostly layout and CSS styling. Any opinion ?

namrataukirde:
Hi @patrixr , You need to write acceptance test cases here. :)

app/controllers/orders/index.js
```diff
@@ -4,5 +4,14 @@ export default searchModule.extend({
 
   searchModelName: "designation",
   unloadAll: true,
-  minSearchTextLength: 2
+  minSearchTextLength: 2,
+
+  onItemLoaded(record) {
+    const orgId = record.get("gcOrganisationId");
+    if (orgId) {
+      this.store.findRecord("gc_organisation", orgId, { reload: false });
```
Darshan-Chauhan: @patrixr , is there any reason for and extra request for loading `gc_organisation` related to Order??
Order response doesn't contain  `gc_organisation`?

patrixr:
> Hi @patrixr , You need to write acceptance test cases here. :)

@namrataukirde I've added some tests. It's a bit incomplete generally as there are quite a few things which are still not working (mock-related issues). I'll create a jira to focus on cleaning up and managing those tests 😄

app/models/designation.js
```diff
@@ -61,6 +62,54 @@ export default Model.extend({
...
+  }),
+
+  transportLabel: Ember.computed("transportKey", function() {
+    const key = this.get('transportKey');
+    return this.get("i18n").t(`order_transports.${key}`);
+  }),
+
+  transportKey: Ember.computed("orderTransport.transportType", function() {
+    const transportType = this.get('orderTransport.transportType');
+    if (transportType === "ggv") {
```
namrataukirde: Can we use switch case here? 🤔

app/models/designation.js
```diff
@@ -61,6 +62,54 @@ export default Model.extend({
...
+    }
+  }),
+
+  transportLabel: Ember.computed("transportKey", function() {
+    const key = this.get('transportKey');
+    return this.get("i18n").t(`order_transports.${key}`);
+  }),
+
+  transportKey: Ember.computed("orderTransport.transportType", function() {
+    const transportType = this.get('orderTransport.transportType');
```
namrataukirde: Good to see a use of const here. :)

Darshan-Chauhan:
@patrixr , too many commits. Can we merge all these commits in one??

patrixr:
> @patrixr , too many commits. Can we merge all these commits in one??

Done

PR #300, opened 2018-10-12
GCW-2155 Fetch designations without associations

patrixr:
Make results on the listing smaller as suggested by Matt. We're hiding the body and keeping the header. 
We're also fetching the designations on that page without any associations to improve performance

### DIFF
--- a/app/templates/orders/order_block.hbs
+++ b/app/templates/orders/order_block.hbs
@@ -67,15 +67,15 @@
         {{/if}}
 
         {{#if (is-or order.designatedOrdersPackages.length order.dispatchedOrdersPackages.length)}}
-            <div class="row order_process_summary">
-              <div class="small-4 columns field">
-                {{ t "order.process_summary" }}
-              </div>
-              <div class="small-8 columns value">
-                <span>{{order.designatedOrdersPackages.length}} <i class="fa fa-shopping-basket"></i></span>
-                <span>{{order.dispatchedOrdersPackages.length}} <i class="fa fa-ship"></i></span>
-              </div>
+          <div class="row order_process_summary">
+            <div class="small-4 columns field">
+              {{ t "order.process_summary" }}
             </div>
+            <div class="small-8 columns value">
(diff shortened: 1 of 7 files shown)


app/templates/orders/order_block.hbs
```diff
@@ -19,64 +19,66 @@
 
     </div>
 
-    <div class="row order_body">
+    {{#unless smallOrderBlocks}}
```
patrixr: The github diff looks a bit messy, but {{#unless smallOrderBlocks}} is the only thing I've added

app/templates/orders/order_block.hbs
```diff
@@ -19,64 +19,66 @@
...
-              <span>{{order.dispatchedOrdersPackages.length}} <i class="fa fa-ship"></i></span>
+              {{order.stockitContact.fullName}}
             </div>
           </div>
-      {{/if}}
+        {{/if}}
 
-    </div>
+        {{#if (is-or order.designatedOrdersPackages.length order.dispatchedOrdersPackages.length)}}
+            <div class="row order_process_summary">
```
Darshan-Chauhan: Indentation

Darshan-Chauhan:
@patrixr , we will need `excludeAssociations` for following lines also
- https://github.com/crossroads/stock.goodcity/blob/df98289615f126c2d4d51e91c31eba1151b238b6/app/routes/index.js#L5
- https://github.com/crossroads/stock.goodcity/blob/df98289615f126c2d4d51e91c31eba1151b238b6/app/components/focus-textfield.js#L43
- https://github.com/crossroads/stock.goodcity/blob/df98289615f126c2d4d51e91c31eba1151b238b6/app/routes/items/partial_designate.js#L14
- https://github.com/crossroads/stock.goodcity/blob/df98289615f126c2d4d51e91c31eba1151b238b6/app/routes/items/search_order.js#L59


Everything else looks fine

patrixr:
@Darshan-Chauhan Updated according to comments above !

PR #308, opened 2018-11-08
GCW 2169 Appointment Quota Settings page

patrixr:
Adding a page to configure the appointment quotas.


app/components/calendar-input.js
```diff
@@ -0,0 +1,74 @@
...
+        monthsFull: moment.months(),
+        monthsShort: moment.monthsShort(),
+        weekdaysShort: moment.weekdaysShort(),
+        disable: _this.get('disableWeekends') === true ? [1,2] : [],
+        clear: false,
+        today: false,
+        close: false,
+        min: moment().toDate(),
+
+        onClose: function() {
```
Darshan-Chauhan: Too many lines in `disInsertElement` callback. Would it be better to separate this function out in some named function outside `didInsertElement` callback?

app/components/calendar-input.js
```diff
@@ -0,0 +1,74 @@
...
+            _this.set("selection", date);
+            setting = true;
+            Ember.run.next(() => {
+              this.set('select', new Date(date), { format: 'ddd mmm d' });
+              setting = false;
+            });
+          }
+        },
+
+        onStart: function(){
```
Darshan-Chauhan: ^

app/components/message-box.js
```diff
@@ -24,14 +24,18 @@ export default Ember.Component.extend({
   actions: {
     btn1Click() {
       if (this.btn1Callback) {
-        this.btn1Callback();
+        if (this.btn1Callback() === false) {
```
Darshan-Chauhan: Are there chances of having null or undefined value for `this.btn1Callback()`

app/components/message-box.js
```diff
@@ -24,14 +24,18 @@ export default Ember.Component.extend({
...
+          return;
+        }
       }
       this.close();
     },
 
     btn2Click() {
       if (this.btn2Callback) {
-        this.btn2Callback();
+        if (this.btn2Callback() === false) {
```
Darshan-Chauhan: ^

app/controllers/appointments.js
```diff
@@ -0,0 +1,276 @@
...
+  isPresetsTab: Ember.computed.equal('selectedTab', 'presets'),
+  isSpecialTab: Ember.computed.not('isPresetsTab'),
+
+  // ---- Data structure
+ 
+  /* Options for the select inputs [00:00, 00:30, ... 23:30]  */
+  timeSlots: Ember.computed(function () {
+    let buildSlot = (hours, minutes) => {
+      let time = this.formatTime(hours, minutes);
+      return Ember.Object.create({ name: time, id: time, time, hours, minutes });
```
Darshan-Chauhan: Would it be possible to direct return hsdh instead of creating an object?

app/controllers/appointments.js
```diff
@@ -0,0 +1,276 @@
...
+      this.set('dateDescription', '');
+    },
+
+    createNewSpecialDate() {
+      if (!this.get('dateDescription') || !this.get('selectedDate')) {
+        this.set('incompleteForm', true);
+        return false;
+      }
+
+      const note = this.get('dateDescription');
```
Darshan-Chauhan: Would it be better yo use `let` here?? As this value will be changed will be every controller object. If it's really a constant than should we move it to `environment.js` file or may be `constants.js`?

Same for all other constants

app/models/role.js
```diff
@@ -1,6 +1,17 @@
...
 import attr from 'ember-data/attr';
+import { hasMany } from 'ember-data/relationships';
 
 export default Model.extend({
-  name: attr('string')
+  name: attr('string'),
+  rolePermissions: hasMany('rolePermission', { async: false }),
+
+  permissions: Ember.computed('rolePermissions.[]', function(){
+    return this.get('rolePermissions').map(rp => rp.get('permission'));
```
Darshan-Chauhan: Ember provides `getEach` options for Array. Would be better to use it here instead of map?

app/models/role.js
```diff
@@ -1,6 +1,17 @@
...
+  }),
+
+  permissionNames: Ember.computed('permissions.[]', function(){
+    return this.get('permissions').map(p => p.get('name'));
```
Darshan-Chauhan: ^

Darshan-Chauhan:
@patrixr , awesome implementation. 👍 
Added some comments

app/components/message-box.js
patrixr: @Darshan-Chauhan the callbacks are initialized with empty methods, no risk

FYI this class is mostly a copy of the calendar for the public holidays selection. There are a few places using this date picker, but they're all slightly different. We should try to consolidate that into a single re-usable component

app/components/calendar-input.js
patrixr: @Darshan-Chauhan I thought about it, but given that there seems to be a mix of scopes (`this` and `_this`) I figured it'd be clearer to leave it as it is.

Do you have any ideas ?
The callbacks need to be bound to the pickadate reference, but also need access to the ember component reference. I could start playing with binds, but I feel that makes it more complicated.

Darshan-Chauhan: You can use `call/apply/bind` for binding `pickadate` reference and pass `_this` as argument to function as component reference. 
I think use of `call` would be better to bind `pickadate` reference.

patrixr: To make it clearer it might be better to pass both of them as arguments I guess
Something like:

```javascript
// separate
function myOnClose(pickadate, component) { 
   // do things
}

// component
{
  onClose: function () {
    // this is bound to pickadate
   // _this is the component
    myOnClose.call(null, this, _this); 
  }
}
```

Darshan-Chauhan: No use of using call if we want to pass first argument as null, might as well use simple function and pass both this as an argument. 

```
// separate
function myOnClose(component) { 
    // here this will be available as pickadate reference.
   // do things
}

// component
{
  onClose: function () {
    // this is bound to pickadate
   // _this is the component
    myOnClose.call(this, _this);  //this bounds pickadate and _this gets passed as an argument
  }
}
```

This should work I think

Darshan-Chauhan: If there are too many arguments to pass then using `apply` would be better option.

PR #331, opened 2018-12-23
GCW 2241 Reschedule orders

patrixr:
Note: there were a lot of hardcoded strings around Stock, so I fix those in a separate PR


Darshan-Chauhan:
Merging this

PR #347, opened 2019-01-14
GCW-2289 Fix order search navigation issues

patrixr:


abhinav098:
hi @patrixr , the PR looks good, but its showing some code-climate issues, can you look into those ?

patrixr:
@abhinav098  modified the ones that seem most relevant

PR #359, opened 2019-01-22
GCW-2321 Clear filter and redirect back to item index page

abhinav098:
Hi Team, this PR fixes the issue where, after clearing the location filter, we weren't getting redirected to items index page. Please review this.
Thanks


app/controllers/item_filters.js
```diff
@@ -54,6 +54,7 @@ export default searchModule.extend({
     clearLocation() {
```
Darshan-Chauhan: Can we change the name to `clearLocationAndRedirect` ?

abhinav098: sure @Darshan-Chauhan

PR #368, opened 2019-02-06
GCW- 2361- Move app version and item create from home screen

abulsayyad123:
Hi Team,
This is PR for JIRA ticket GCW-2361-Move app version and item create from home screen.  This PR includes:
* `Add inventory Item` is moved from dashboard to Menu and Item page.
* App version is moved from dashboard to Menu top (and it is available only for mobile).
* Changed and fixed test cases accordingly.
* Fixed test cases error related to mocking of `/summary` and `/booking_type` `GET` resource.
Thanks.

### DIFF
--- a/app/controllers/search_code.js
+++ b/app/controllers/search_code.js
@@ -16,6 +16,7 @@ export default Ember.Controller.extend({
   searchPlaceholder: t("search.placeholder"),
   i18n: Ember.inject.service(),
   isSearchCodePreviousRoute: Ember.computed.localStorage(),
+  previousPath: null,
 
   allPackageTypes: Ember.computed("fetchMoreResult", function(){
     return this.store.peekAll('code').filterBy('visibleInSelects', true);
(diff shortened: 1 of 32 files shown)


app/controllers/search_code.js
```diff
@@ -106,12 +107,15 @@ export default Ember.Controller.extend({
     },
 
     cancelSearch() {
+      let previousPath = this.get('previousPath');
       Ember.$("#searchText").blur();
       this.send("clearSearch", true);
       if(this.get("backToNewItem")) {
         this.transitionToRoute("items.new");
+      } else if (previousPath) {
+        this.transitionToRoute(previousPath);
```
Darshan-Chauhan: Can we only keep history.back()?
As it is more reliable and will also take to previous path and also less code in route to get the name of previous path

Darshan-Chauhan:
@abulsayyad123 , just for future reference, you can add mockjax mocks to a variable and clear them after each test case is run.
So you won't get any issues of single test cases passing and same test cases failing in group. :)

app/controllers/search_code.js
abulsayyad123: I wasn't very sure about `history.back()`, that is why I wrote `previousPath` logic. Just to be on safer side I kept `history.back()` as a last option. If `history.back()` is more reliable than we can go for it. 👍

Darshan-Chauhan: Yes, I think it would better :)

app/controllers/search_code.js
```diff
@@ -111,7 +111,7 @@ export default Ember.Controller.extend({
       if(this.get("backToNewItem")) {
         this.transitionToRoute("items.new");
       } else {
-        this.transitionToRoute("index");
+        window.history.back();
```
namrataukirde: :+1:

PR #374, opened 2019-02-13
GCW-2335 basic setup chat

abhinav098:
Hi team this PR contains changes for stock app for chat between orders.


app/controllers/subscription.js
```diff
@@ -205,13 +207,18 @@ export default Ember.Controller.extend({
...
         payload[type] = item;
         this.store.pushPayload(payload);
     } else if (existingItem) { //delete
       this.store.unloadRecord(existingItem);
     }
     run(success);
+
+    // mark message as read if message will appear in current view
+    if (type === "message") {
+      console.log('mark messages read here');
```
patrixr: Something missing ? :)

app/styles/templates/orders/_messages.scss
```diff
@@ -0,0 +1,78 @@
+body{
```
patrixr: couple of things : 

It's important to scope all the css rules to the page you're building, so make sure to put all those rules under a class/ID of your page. It limits the risks.

e.g
```
.order-chat-page {
 // all rules here
}
```

right now it's in the global scope, meaning if on some completely different page I use a similar class (e.g `message-section` is fairly generic), it would apply styles that I don't want.

Another thing is the `body { }` rule. That's quite dangerous cause it'll apply it for the whole app (potentially unwanted), better to have that rule on the container div of you page

app/controllers/subscription.js
abhinav098: yes @patrixr , mark read feature in this chat is pending, I'll remove this code for now.

app/styles/templates/orders/_messages.scss
abhinav098: Yeah you're right. I'll do that. Thanks @patrixr

app/models/message.js
```diff
@@ -0,0 +1,38 @@
...
+  state: attr('string', {
+    defaultValue: 'read'
+  }),
+  sender: belongsTo('user', {
+    async: true
+  }),
+  designation: belongsTo('designation', { async: false }),
+
+  myMessage: Ember.computed(function () {
+    var session = getOwner(this).lookup("service:session");
```
patrixr: Couldn't the service simply be added to the Model ? 
e.g 
```
export default DS.Model.extend({
    session: Ember.inject.service()
    //...
})
```
and then use `this.get('session')`

I'm no  Ember expert, but `getOwner(this).lookup` **_feels_** a bit hackish 🤔

patrixr:
Looking good @abhinav098  : )

app/models/message.js
Darshan-Chauhan: `session` service should be accessible everywhere as we've inject it using initializers

app/routes/orders/conversation.js
```diff
@@ -1,3 +1,21 @@
 import detail from './detail';
+import Ember from 'ember';
 
-export default detail.extend();
+export default detail.extend({
+  model(params){
+    var designation = this.store.peekRecord("designation", params.order_id, {
+      reload: true }) || this.store.findRecord('designation', params.order_id);
+    return Ember.RSVP.hash({
+      designation: designation,
```
Darshan-Chauhan: This won't work in case if it fetches designation from API.  As it's asynchronous . 
You'll need nested promise

app/models/message.js
patrixr: @Darshan-Chauhan meaning we can do `this.get('session')` without injecting it ?

app/routes/orders/conversation.js
patrixr: I think replacing it with `RSVP.all` should do the trick. It'll resolve the promises, and those that aren't promises will just pass through

app/models/message.js
Darshan-Chauhan: @patrixr , yes. Look at this initializer => https://github.com/crossroads/stock.goodcity/blob/master/app/initializers/session.js

patrixr: aha, good to know :)

abhinav098: yes, that would be right, thanks @Darshan-Chauhan @patrixr

app/controllers/subscription.js
```diff
@@ -195,7 +195,9 @@ export default Ember.Controller.extend({
       return false;
     }
 
-    this.setFavImage(item, data, type);
+    if(type.toLowerCase() !== "message"){
+      this.setFavImage(item, data, type);
```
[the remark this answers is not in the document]
abhinav098: @Darshan-Chauhan , reason for adding this condition was, not to setFavImage if type is message.

app/components/online-button.js
```diff
@@ -0,0 +1,32 @@
+import Ember from "ember";
+
+export default Ember.Component.extend(Ember.TargetActionSupport, {
+  tagName: "div",
```
abulsayyad123: No need to specify `div` by default components are div.

app/components/variable-height-textarea.js
```diff
@@ -0,0 +1,37 @@
...
+          if (_this.get("value") !== "") {
+            window.scrollTo(0, document.body.scrollHeight);
+          }
+        } else {
+          Ember.$(textarea)
+            .css({
+              'height': 'auto',
+              'overflow-y': 'auto'
+            })
+            .height(105);
```
abulsayyad123: You are specifying height in CSS and also adding separately ?

app/controllers/orders/conversation.js
```diff
@@ -1,3 +1,76 @@
...
+  groupedMessages: Ember.computed("sortedMessages", function () {
+    return this.groupBy(this.get("sortedMessages"), "createdDate");
+  }),
+
+  groupBy: function (content, key) {
+    var result = [];
+    var object, value;
+
+    content.forEach(function (item) {
+      value = item.get ? item.get(key) : item[key];
```
abulsayyad123: No need for this condition , you can do simply by `Ember.get(item, key)`

app/controllers/subscription.js
```diff
@@ -205,13 +207,18 @@ export default Ember.Controller.extend({
       return;
     }
 
-    if (["create","update"].contains(data.operation)) {
+    if (["create","update"].includes(data.operation)) {
```
abulsayyad123: I think `indexOf` should be used here, for compatibility with I.E 😆

app/routes/orders/conversation.js
abulsayyad123: I think adding this fetching `designation` directly inside the `RSVP.hash` should work , it will resolve all request altogether ?

app/components/online-button.js
abhinav098: sure, ill remove it.

app/controllers/orders/conversation.js
abhinav098: sure, thanks @abulsayyad123

app/controllers/subscription.js
abhinav098: i changed it to `includes` due to console warning. should i use `indexOf` or change it back to `contains` ?

abulsayyad123: Use `indexOf`, considering I.E.

PR #375, opened 2019-02-14
GCW-2366 logistics tab : Add rows

patrixr:
Adds extra lines of data to the logistic tabs. Those are not yet editable.

From the design :
screenshot 2019-02-14 at 11 40 42 am

### DIFF
--- a/app/controllers/orders/order_types.js
+++ b/app/controllers/orders/order_types.js
@@ -1,3 +1,54 @@
 import detail from "./detail";
+import Ember  from 'ember';
+import _ from 'lodash';
 
-export default detail.extend();
+export default detail.extend({
+
+  logisticDataRows: Ember.computed('model', function () {
+    const order = this.get('model');
+    return [
+      // Schedule
+      {
+        label: this.get('i18n').t('order_details.logistics.scheduled'),
+        text: moment.tz(order.get('orderTransport.scheduledAt'), "Asia/Hong_Kong").format('dddd Do MMMM hh:mm a'),
+        action: 'openSchedulePopup',
+        icon: 'clock',
+        name: 'reschedule'
+      },
+      // Type
+      {
+        label: this.get('i18n').t('order_details.logistics.type'),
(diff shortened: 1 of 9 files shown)


app/controllers/orders/order_types.js
```diff
@@ -1,3 +1,54 @@
...
 
-export default detail.extend();
+export default detail.extend({
+
+  logisticDataRows: Ember.computed('model', function () {
+    const order = this.get('model');
+    return [
+      // Schedule
+      {
+        label: this.get('i18n').t('order_details.logistics.scheduled'),
```
Darshan-Chauhan: This seems like repetition of same... Can we create a function which will take params and returns hash depending on param??

patrixr: we can, but it'd end up similar, calling the function again and again.

e.g
```
makeRow(
        this.get('i18n').t('order_details.logistics.type'),
        order.get('bookingTypeLabel'),
        function action() {},
        'tv',
        'type'
)
```

I'm okay with that, but i don't find it significantly better.
It's just data after all.
We could move everything to the `hbs` and not have a foreach, but i think it clutters the template too much, this is long but very understandable

patrixr: I could make separate functions, e.g `makeVehicleRow()` , `makeScheduleRow()`, and return 
```
[ makeVehicleRow(), makeScheduleRow() ]
```

but then again, is it really a better alternative, i'm not so sure tbh

Darshan-Chauhan:
@patrixr , don't merge anything to master yet.
I'm working on fix for language-switcher which we want to release. 
Once the fix is released then feel free to merge. :)

patrixr:
> @patrixr , don't merge anything to master yet.
> I'm working on fix for language-switcher which we want to release.
> Once the fix is released then feel free to merge. :)

Got it :)

Darshan-Chauhan:
@patrixr , feel free to merge now :)

PR #377, opened 2019-02-15
Order details : Make logistic fields editable

patrixr:


patrixr:
@Darshan-Chauhan This is the next step of the PR you last reviewed. The different objects were starting to look different, so I made separate functions for each of them.

PR #378, opened 2019-02-18
Gcw 2365 expand capabilities of order summary tab

Darshan-Chauhan:


app/controllers/orders/client_summary.js
```diff
@@ -0,0 +1,35 @@
...
+
+export default detail.extend({
+  showBeneficiaryModal: false,
+
+  actions: {
+    removeBeneficiaryModal() {
+      this.toggleProperty('showBeneficiaryModal');
+    },
+
+    deleteBeneficiary() {
```
patrixr: note : the `detail` base controller extends the `GoodcityController` which has a `deleteRecords`
You could replace a chunk of this code by `this.deleteRecords([ beneficiary ])`
It also handles the spinner (aka loadingView) 😄

app/controllers/orders/client_summary.js
```diff
@@ -0,0 +1,35 @@
...
+      const beneficiaryId = this.get("model.beneficiaryId");
+      var url = `/beneficiaries/${beneficiaryId}`;
+      var beneficiary = this.get("store").peekRecord("beneficiary", beneficiaryId);
+      if(beneficiary) {
+        var loadingView = getOwner(this).lookup('component:loading').append();
+        new AjaxPromise(url, "DELETE", this.get('session.authToken'))
+          .then(data => {
+            this.get("store").pushPayload(data);
+            new AjaxPromise(`/orders/${this.get("model.id")}`, "PUT", this.get('session.authToken'), { order: { beneficiary_id: null } })
```
patrixr: This new promise needs to be returned, otherwise the `finally` hook will run before the promise is truly completed

app/models/user.js
```diff
@@ -23,5 +26,19 @@ export default Addressable.extend({
...
   organisationsUsers: hasMany('organisationsUsers', {async: false}),
 
-  // image:       belongsTo('image', { async: false }),
+  position: Ember.computed('organisationsUsers.[]', function() {
+    const organisationsUsers = this.get("organisationsUsers");
+    return organisationsUsers.length ? 
+    organisationsUsers.get("firstObject.position") : '';
+  }),
+
+  onlineStatus: Ember.computed('lastConnected', 'lastDisconnected', function() {
```
patrixr: Maybe the returned value should be coming from an enum 
e.g 
```javascript
const CONN_STATUSES = {
    ONLINE: 'online',
    OFFLINE: 'offline',
    UNKNOWN: 'unknown'
}
```
and then can be interpolated to build the translation string `this.get('i18n').t('connection.status.${status}')`

the mix of boolean and strings is a bit weird :D

app/components/time-ago.js
```diff
@@ -0,0 +1,13 @@
+import Ember from 'ember';
```
patrixr: could be a helper, i think there's already a `display-datetime` which could be copied/adapted

app/templates/orders/_summary_tabs.hbs
```diff
@@ -0,0 +1,19 @@
+<section class="main-section order_item_list_tabs">
+  <div class="row">
+    <div class="small-12 columns tab_row">
+      <dl class="tabs" data-tab>
+        {{#link-to 'orders.contact_summary' model.id tagName="dd" classNames="small-6 medium-6 columns text-center"}}
+          {{#link-to 'orders.contact_summary' model.id}}
+            <div>Contact</div>
```
patrixr: btw, Matt asked me to go through Stock and make sure everything is in the translation file. It'll take a while, but if you find any strings hardcoded into the templates, make sure to move them to the localisation files 😄

app/templates/orders/client_summary.hbs
```diff
@@ -0,0 +1,124 @@
...
+<section class="main-section order_page">
+  {{partial "orders/order_status_bar"}}
+  {{partial "orders/order_tabs"}}
+  {{partial "orders/summary_tabs"}}
+  <div class="row">
+    <div class="small-12 columns main_details">
+      {{#if model.beneficiary}}
+        <div class="row">
+          <div class="small-5 columns">
+            Remove Client
```
patrixr: Same as comment above :)

app/controllers/orders/client_summary.js
Darshan-Chauhan: Thanks, I'll make changes :)

app/controllers/orders/client_summary.js
Darshan-Chauhan: Oh, I forgot!

app/models/user.js
Darshan-Chauhan: Good idea.

app/components/time-ago.js
Darshan-Chauhan: Yeah, i felt the same way. 
The logic was copied from Admin app

app/templates/orders/_summary_tabs.hbs
Darshan-Chauhan: I'm still not used to adding translations in stock app. 😄 
I'll make the chagnes

app/controllers/orders/client_summary.js
Darshan-Chauhan: Although that detail controller is for `orders` route and this is for `order`, I just noticed.

app/models/user.js
Darshan-Chauhan: Will still need the conditions for showing online/not-connected status. So I think it'll the same thing. (Also instead of offline we show `not connected`).
I did added translations though

patrixr: Conditions are fine, it's just that the return value is very hard to utilise anywhere else if needed :)

app/templates/orders/_summary_tabs.hbs
patrixr: Same, doing that now myself

app/controllers/orders/client_summary.js
patrixr: Well, if you feel it's useful feel free to have your controller extend the `GoodcityController`, and feel free to add any handy/reusable code there as well : )

app/models/user.js
patrixr: @Darshan-Chauhan maybe we can start a practice to name the properties that are translations with a `Label` suffix ?

e.g `onlineStatusLabel` 

That way whenever we see a name similar to this, we know that it's a translation string.
To avoid anyone doing something like `if (user.get('onlineStatus') === 'Not connected)`  (which is dangerous)
Just an idea

app/controllers/orders/client_summary.js
Darshan-Chauhan: I was also thinking about using `axios` instead of AjaxPromise

patrixr:
good work ! left a few comments, looks okay to be merged apart from that :)

app/models/user.js
Darshan-Chauhan: I liked the idea 👍

PR #379, opened 2019-02-20
GCW-2366 Add sticky note to Logistics tab

patrixr:


app/components/auto-resize-textarea.js
```diff
@@ -29,5 +29,25 @@ export default Ember.TextArea.extend({
         .css({'height':'auto','overflow-y':'auto'})
         .height(105);
     }
+  },
+
+  callAction(action, data = null) {
+    if (typeof action === 'function') {
+      return action(data);
+    } else if (typeof action === 'string') {
+      this.sendAction(action, data);
```
Darshan-Chauhan: I feel that component and controller logic should be separated  so that it would be easier to handle.
Would it be better to write actions in component rather than in controller??

patrixr: They are already separate. That's just a helper function. I pass in the action from the controller as a parameter and the component calls it when the event triggers

PR #380, opened 2019-02-20
GCW-2365 Inline edit for beneficiary and order

Darshan-Chauhan:
Hi Team,

This PR includes inline edit feature for Order and beneficiary.

Please review.
Thanks

### DIFF
--- a/app/controllers/orders/client_summary.js
+++ b/app/controllers/orders/client_summary.js
@@ -6,6 +6,22 @@ const { getOwner } = Ember;
 export default detail.extend({
   showBeneficiaryModal: false,
 
+  titles: Ember.computed(function(){
+    return [
+      { name: "Mr", id: "Mr" },
+      { name: "Mrs", id: "Mrs" },
+      { name: "Miss", id: "Miss" },
+      { name: "Ms", id: "Ms" }
+    ];
+  }),
+
+  identityTypes: Ember.computed(function() {
+    return [
+      { name: 'Hong Kong Identity Card', id: 1 },
+      { name: 'Asylum Seeker Recognizance Form', id: 2}
+    ];
+  }),
+
   actions: {
     removeBeneficiaryModal() {
       this.toggleProperty('showBeneficiaryModal');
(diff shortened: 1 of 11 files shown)


Darshan-Chauhan:
Regarding code climate issues, I've added separate file for for each inline edit field so that it would be easier to handle in future. So please ignore.

app/controllers/orders/client_summary.js
```diff
@@ -6,6 +6,22 @@ const { getOwner } = Ember;
...
+      { name: "Mr", id: "Mr" },
+      { name: "Mrs", id: "Mrs" },
+      { name: "Miss", id: "Miss" },
+      { name: "Ms", id: "Ms" }
+    ];
+  }),
+
+  identityTypes: Ember.computed(function() {
+    return [
+      { name: 'Hong Kong Identity Card', id: 1 },
```
namrataukirde: I believe for identity type we have records in db. I am not sure we are preloading those or not in ember-data but I feel instead of hardcoded values picking those from ember-data will be better. Let me know your thoughts. :)

Darshan-Chauhan: We only load `booking_type` in preload and I think hardcoded value is okay for select list.

app/controllers/orders/client_summary.js
```diff
@@ -6,6 +6,22 @@ const { getOwner } = Ember;
...
+  titles: Ember.computed(function(){
+    return [
+      { name: "Mr", id: "Mr" },
+      { name: "Mrs", id: "Mrs" },
+      { name: "Miss", id: "Miss" },
+      { name: "Ms", id: "Ms" }
+    ];
+  }),
+
+  identityTypes: Ember.computed(function() {
```
abulsayyad123: Would it be better to add model and fetch `identityType`?  Just a Thought.

Darshan-Chauhan: It will only return one identityType and we need both identityType for Select list so  :)

app/controllers/orders/client_summary.js
namrataukirde: Can we preload identity-type? What if someone deletes and recreates records then Id's will change I think. So better to send ids of the actual record. Any thoughts?

We previously did same with booking_type then did it as per actual records.

If you still think it's fine then keep it. 👍

abhinav098: I second @namrataukirde on this.

PR #381, opened 2019-02-22
GCW-2335 fix sender issue chat

abhinav098:


namrataukirde:
Check on monday how we can keep async false and load it actually from api.

abhinav098:
sure @namrataukirde

PR #393, opened 2019-03-01
GCW-2414 read unread tracking of messages

abhinav098:
Hello Team this PR is for read unread tracking of messages.
Please review.

### DIFF
--- a/app/controllers/orders/conversation.js
+++ b/app/controllers/orders/conversation.js
@@ -3,6 +3,7 @@ import config from '../../config/environment';
 import detail from './detail';
 
 export default detail.extend({
+  messagesUtil: Ember.inject.service("messages"),
   body: '',
   isPrivate: false,
   backLinkPath: "",
@@ -75,6 +76,11 @@ export default detail.extend({
 
       // Animate and scroll to bottom
       this.autoScroll();
+    },
+    markRead(){
+      this.get("sortedMessages")
+        .filterBy('state', 'unread')
+        .forEach(m => this.get("messagesUtil").markRead(m));
     }
   }
 });
--- a/app/controllers/orders/conversation.js
+++ b/app/controllers/orders/conversation.js
@@ -80,7 +80,7 @@ export default detail.extend({
     markRead(){
       this.get("sortedMessages")
         .filterBy('state', 'unread')
-        .forEach(m => this.get("messagesUtil").markRead(m));
+        .forEach(message => this.get("messagesUtil").markRead(message));
     }
   }
 });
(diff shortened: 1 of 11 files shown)


app/controllers/orders/conversation.js
```diff
@@ -75,6 +76,11 @@ export default detail.extend({
 
       // Animate and scroll to bottom
       this.autoScroll();
+    },
+    markRead(){
+      this.get("sortedMessages")
+        .filterBy('state', 'unread')
+        .forEach(m => this.get("messagesUtil").markRead(m));
```
abulsayyad123: use more expressive variable name `m` doesn't say anything.

abhinav098: sure, i'll name it `message`

app/controllers/subscription.js
```diff
@@ -214,6 +215,39 @@ export default Ember.Controller.extend({
     } else if (existingItem) { //delete
       this.store.unloadRecord(existingItem);
     }
+
+    // mark message read here
+    if (type === "message") {
+      var router = this.get("target");
+      var currentUrl = window.location.href.split("#").get("lastObject");
```
abulsayyad123: What type of url we are getting here ? I think it would be better if we can fetch this in route and make it available in controller. Thoughts ?
CC: @namrataukirde

abhinav098: This will return `http://localhost:4202/orders/21191/conversation`, we need this to verify that user is on messages page. and then mark the latest message `read`.

abhinav098: This scenario is, when the user is chatting with stock user or vice versa. New message which has come up, we need to mark that message `read` . 
So i have written code in a way that it verifies the currentHref, with the generated messageUrl.
and if they match, i hit the mark_read method from the messages_controller.

abhinav098: > I think it would be better if we can fetch this in route

That wont work for new incoming messages if user is still chatting.

app/services/messages.js
```diff
@@ -0,0 +1,30 @@
+import Ember from "ember";
+const { getOwner } = Ember;
+
+export default Ember.Service.extend({
+  logger: Ember.inject.service(),
+  session: Ember.inject.service(),
+
+  markRead: function (message) {
```
steveyken: as discussed on slack, it would be good to look at a way to push a list of messages to mark read rather than lots of inidividual 'mark_read' calls.

abhinav098: yep, right but @steveyken, for that we'll need to make changes to all the 4 apps accordingly. As currently on `donor` and `admin` also the messages are marked `read` in similar fashion.
Also the api would also require changes, as currently it is designed to handle the marking of messages to `read` for a single instance of message.

Let me know should I go ahead with the changes? 

cc - @namrataukirde

steveyken: no need to change now. let's come back to it.

PR #400, opened 2019-03-06
GCW-2425 fix chat issues

abhinav098:
- restrict blank message


app/components/variable-height-textarea.js
```diff
@@ -3,6 +3,7 @@ import Ember from 'ember';
 export default Ember.TextArea.extend({
   tagName: "textarea",
   attributeBindings: ["disabled"],
+  classNames: 'message_textbar',
```
Darshan-Chauhan: Follow hyphenated words for class name

abhinav098: sure @Darshan-Chauhan , i'll use hyphenated class name.

PR #401, opened 2019-03-06
FIX:  Validation for people helped in order summary

abulsayyad123:
Hi Team,
This PR is regarding adding validation for People Helped input in Summary tab.
Thanks

### DIFF
--- a/app/components/order-inline-number-input.js
+++ b/app/components/order-inline-number-input.js
@@ -5,10 +5,15 @@ const { getOwner } = Ember;
 
 export default BeneficiaryInlineInput.extend({
   type: "tel",
+  orderCopy: Ember.computed.readOnly('order'),
 
   whichKey(e, key) {
     var keyList = [13, 8, 9, 39, 46, 32];
     return e.ctrlKey && key === 86 || keyList.indexOf(key) >= 0 || key >= 48 && key <= 57;
+  },
+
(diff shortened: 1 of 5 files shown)


app/components/order-inline-number-input.js
```diff
@@ -5,22 +5,28 @@ const { getOwner } = Ember;
 
 export default BeneficiaryInlineInput.extend({
   type: "tel",
+  orderCopy: Ember.computed.readOnly('order'),
```
Darshan-Chauhan: What is the use of readOnly here? Why it is needed?

abulsayyad123: It is to avoid updating the `order` object and maintain the `order` object when are coming back to the tab.

Darshan-Chauhan: I don't see any manipulation of Order in the file or PR. 
I don't think this is necessary.

abulsayyad123: When you delete value here (https://github.com/crossroads/stock.goodcity/pull/401/files#diff-f26ff51c81035b6e0e71a7ceb8f18e1dR28), and you change tab and come back and add back value with `peopleHelped` which was empty earlier (because of reference). So ,  to maintain that state and make it unchangeable it was required to make a readonly.

PR #402, opened 2019-03-08
GCW- 2421 add total request count

abulsayyad123:
Hi Team,
This PR is regarding jira ticket GCW-2421- add total request count on summary tab(https://jira.crossroads.org.hk/browse/GCW-2421). Please review the PR and let me know if any change is required.

Thanks.


app/styles/templates/orders/_detail.scss
```diff
@@ -514,6 +514,11 @@
     .arrow-icon {
       float: right;
     }
+
+    .state-count {
+      margin-left: 2%;
```
Darshan-Chauhan: you can just write `margin: 0 2%`

app/routes/orders/detail.js
```diff
@@ -65,9 +66,21 @@ export default getOrderRoute.extend({
...
+      "GET",
+      this.get("session.authToken")
+    );
+    Object.assign(result.designation, result.meta.counts);
+    Ember.run(function() {
+      _this.store.pushPayload(result);
+    });
     return this.loadIfAbsent("designation", params.order_id);
   },
+  /* jshint ignore:end */
```
patrixr: Couple things regarding the piece of code above :

- Is `Ember.run(function() { ... })` needed ?
- Rather than assign the counts to the designation, could we maybe have it as a separate data set ?

e.g 
```
return RSVP.hash({
    orderCounts: result.meta.counts,
    designation: result.designation
})
```

- This line : `this.loadIfAbsent("designation", params.order_id)` becomes useless now because you're fetching it regardless.

Generally we don't want to load from the api every time, because it will trigger even if the data is already here (e.g when we switch tabs)
I'm actually modifying the same file with some caching mechanisms : https://github.com/crossroads/stock.goodcity/blob/filters-refactor/app/routes/orders/detail.js#L74

Lemme know what you think, we can sync and have a compromise

abulsayyad123: Ember data request i.e findRecord is not fetching metadata info. That is why I had to make `ajaxPromise` request. I am giving a try to above code.

patrixr: I think it's fine to make an Ajax promise (the other option would be to create a `meta` model)
I just don't think modifying the record is needed, better send the counts to the controller as a separate variable

abulsayyad123: Returning 
```
return RSVP.hash({
    orderCounts: result.meta.counts,
    designation: result.designation
})
```
will require to make changes in lots of places because all routes, controllers and views under `orders` nested routes are accessing data using `model.`. If we are returning both `designation` and `counts` will require to make changes in lots of places. So, is it fine to make all those changes ?

ping: @patrixr

patrixr: I would lean towards making the change. It'll be useful in the long run.
It's mostly a renaming, so the side effects should be limited

As a practice, we should probably get used to always returning a hash as a model, even if we only have a single item.

abulsayyad123:
Closing this PR due to changes in the approach. This code is not compatible with the latest changes, creating new PR for this.

PR #407, opened 2019-03-11
GCW-2425 fix chat issues

abhinav098:
Hi team 
This PR fixes the issue of first message issue mentioned in the link below, also this PR contains the changes for removal of sender model.
https://jira.crossroads.org.hk/browse/GCW-2425?focusedCommentId=33313&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-33313
Thanks

### DIFF
--- a/app/models/message.js
+++ b/app/models/message.js
@@ -18,7 +18,7 @@ export default DS.Model.extend({
   }),
   designation: belongsTo("designation", { async: false }),
 
-  myMessage: Ember.computed(function() {
+  myMessage: Ember.computed("sender", function() {
     return this.get("sender.id") === this.get("session.currentUser.id");
   }),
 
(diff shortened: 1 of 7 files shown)


app/models/message.js
```diff
@@ -1,35 +1,35 @@
...
-  sender: belongsTo('sender', {
+  sender: belongsTo("user", {
     async: false
   }),
-  designation: belongsTo('designation', { async: false }),
+  designation: belongsTo("designation", { async: false }),
 
-  myMessage: Ember.computed(function () {
-    return this.get("sender.id") === this.get('session.currentUser.id');
+  myMessage: Ember.computed(function() {
```
abulsayyad123: You are returning values on `sender` property, then you should add `sender` on the CP values list.
```
Ember.computed('sender', function(){})
```
Add similar change for other CP methods too.

abhinav098: sure, thanks

PR #412, opened 2019-03-13
GCW - 2473 request populate on first attempt

abulsayyad123:
Hi Team, 
This PR is for JIRA ticket GCW-GCW-2473- Request goods not populating on first attempt.(https://jira.crossroads.org.hk/browse/GCW-2473).  I have also added change for transport details on reflecting not reflecting after changing on confirm page. 
Thanks.
Changes:
https://github.com/crossroads/stock.goodcity/pull/412/files#diff-530ad3f505e5fb3188441a432509c4f5R16
https://github.com/crossroads/stock.goodcity/pull/412/files#diff-530ad3f505e5fb3188441a432509c4f5R32
https://github.com/crossroads/stock.goodcity/pull/412/files#diff-34c3f50ad8fe4ce77a6341f846c73786R7
https://github.com/crossroads/stock.goodcity/pull/412/files#diff-272caf54bdda0689a6347736642ec50eR163
https://github.com/crossroads/stock.goodcity/pull/412/files#diff-272caf54bdda0689a6347736642ec50eR145


app/components/add-request.js
```diff
@@ -1,43 +1,61 @@
...
+    }
   }),
 
   actions: {
     deleteRequest(reqId) {
       var i18n = this.get("i18n");
       this.get("messageBox").custom(
-        i18n.t("order.request.remove_req").string + this.get("model.code"),
-        i18n.t("order.request.remove").string, () => this.send("removeRequest", reqId),
+        i18n.t("order.request.remove_req").string + this.get("order.code"),
```
patrixr: just a thought, I think `i18n` supports string interpolation (not sure though).
It'd be worth testing if we can do something like:
```
I18n.translations = {
  en: {
    hello: 'hello {{name}}',
    }
  }
}

// hbs
{{ t 'hello' name=myName }}

```

app/models/designation.js
```diff
@@ -135,46 +142,56 @@ export default Model.extend({
     }
   }),
 
-  appointmentTransport: Ember.computed("orderTransport", function() {
-    let i18n = this.get("i18n");
-    return this.get("orderTransport.transportType") === "self" ?
-      i18n.t("order.appointment.self_vehicle") : i18n.t("order.appointment.hire_vehicle");
-  }),
+  appointmentTransport: Ember.computed(
```
patrixr: Unrelated to your changes, but I'm thinking we should rename properties which are translations, to something like `appointmentTransportLabel`, to make it crystal clear that it's a display string

abulsayyad123: Sure will do.

app/components/add-request.js
Darshan-Chauhan: @patrixr , yes it does support it.

CC: @abulsayyad123

abulsayyad123: Yes it is possible, and it has implemented where we need to pass attributes to translations. In this case, message-box is used at many places which require to add changes at many place. I think we can take away this point for future reference.

Darshan-Chauhan: It's not change in structure of messageBox.

PR #413, opened 2019-03-13
GCW-2475 fix blank notification issue

abhinav098:
Hello team this PR fixes the issues for blank notification on stock app.

main changes 
https://github.com/crossroads/stock.goodcity/pull/413/files#diff-1616934d9aeb74b7bbf71eff7a6ed2a2R31

https://github.com/crossroads/stock.goodcity/pull/413/files#diff-d9a8e6f8851e76cab05b106d305d825dR8

https://github.com/crossroads/stock.goodcity/pull/413/files#diff-31a61ce2fb36709b87beea82d00933bfR92

Screen shot for notification
Screen Shot 2019-03-13 at 2 46 18 PM

### DIFF
--- a/app/controllers/notifications.js
+++ b/app/controllers/notifications.js
@@ -10,7 +10,7 @@ export default Ember.Controller.extend({
     }
   }),
 
-  nextNotification: Ember.computed('model.[]', function() {
+  nextNotification: Ember.computed("model.[]", function() {
     //retrieveNotification is not implemented here because it needs to call itself
     return this.retrieveNotification();
   }),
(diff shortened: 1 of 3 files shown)


app/controllers/notifications.js
```diff
@@ -27,23 +27,29 @@ export default Ember.Controller.extend({
     return notification;
   },
 
-   setRoute: function(notification) {
-      notification.route = ['orders.detail', notification.order_id];
-   },
+  setRoute: function(notification) {
+    if (notification.category === "message") {
+      return (notification.route = [
```
patrixr: not a huge fan of returning an assignment personally

abhinav098: We'll have to return from this block if the condition passes. What i can do is, I can return nothing after assigning the `notification.route`. Is that fine @patrixr ?
Maybe something like 
```if (true) {
a = xyz();
return;
}```

PR #419, opened 2019-03-22
Refactor subscription.js

patrixr:
Hi team,

As I was adding live update support for appointment_slots, @steveyken and I found quite a few problems regarding those. Both on the server side and front-end.
This PR is meant to fix the architectural issues of `subscription.js`

Changes : 
- Subscription is now a `service`, as opposed to a `controller` (which it never should have been)
- Removal of all the dom manipulation code from there (I had reviewed that but never noticed it was in this file, my bad)
- Removal of the injected controllers (was meant to trigger changes)
- Subscription is now an event emitter, to which the other controllers can register and therefore run the custom dom related code within
- Removal of the multitude of `if/else` statements which was cluttering the file - Replaced by configurations for each type
- Fix of a bug where the `delete` operation would still trigger a `findRecord` which would cause a popup coming out of nowhere
- Moved code snippets around to helper functions with meaningful names to make the file more readable
- Added some calls to `console.warn` in a few places, the idea behind that is to understand what data isn't being used, and update the api to not send it at all
- We are now sending the device_id with every request, that allows us to detect which live updates originate from ourselves, and therefore not process those

**notes:** 

- A lot of the modification of `items` was made irrelevant by the fact that that they were being reloaded from the server. We also found that the backend was not actually firing updates for items anyways (aka packages), a fix for that is coming soon

- This change might cause some slight side effects in the short term, but it'll be a more solid base to later improve and build on top of


abhinav098:
@patrixr could you please check these merge conflicts ?

PR #421, opened 2019-04-01
 Itunes Fix- goto not available

Darshan-Chauhan:
Hi Team,

This PR contains fix for iTunes: GoTo not available.
I've just commented the code in case we want to enable this feature in future.

Please review.
Thanks

### DIFF
--- a/app/templates/sticky_footer.hbs
+++ b/app/templates/sticky_footer.hbs
@@ -2,7 +2,7 @@
   <div class="navigation_menus">
     <div class="row">
       <div class="small-12 columns">
-        <ul class="small-block-grid-5 small-block-grid-5 list">
+        <ul class="small-block-grid-4 list">
           {{#link-to 'index' tagName='li' href=false}}
             <i>{{fa-icon 'home'}}</i>
             {{t "index.home"}}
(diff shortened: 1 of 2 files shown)


app/templates/sticky_footer.hbs
```diff
@@ -22,7 +22,7 @@
             <i>{{fa-icon 'cube'}}</i>{{t "index.items"}}
           {{/link-to}}
 
-          <li><i>{{fa-icon 'qrcode'}}</i>{{t "index.goto"}}</li>
+          {{!-- <li><i>{{fa-icon 'qrcode'}}</i>{{t "index.goto"}}</li> --}}
```
abhinav098: Why have we commented out this line ?

Darshan-Chauhan: Because it doesn't redirect anywhere

abhinav098: oh, okkay. So Is that's why the ios-build was failing ?

Darshan-Chauhan: No, that's why they rejected our app on iTunes

abhinav098: Okay great.. excellent work 👍

namrataukirde:
👍

PR #423, opened 2019-04-02
GCW-2394 restrict partial dispatch

abhinav098:
Hello team, 
This PR restricts user to partial dispatch if order is not processed.
Please review and comment
Thanks.

### DIFF
--- a/app/controllers/items/partial_undesignate.js
+++ b/app/controllers/items/partial_undesignate.js
@@ -1,8 +1,10 @@
 import Ember from "ember";
-import AjaxPromise from 'stock/utils/ajax-promise';
+import AjaxPromise from "stock/utils/ajax-promise";
 const { getOwner } = Ember;
 
 export default Ember.Controller.extend({
+  messageBox: Ember.inject.service(),
+  i18n: Ember.inject.service(),
   promptUserPopUp: false,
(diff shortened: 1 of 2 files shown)


app/controllers/items/partial_undesignate.js
```diff
@@ -13,57 +15,88 @@ export default Ember.Controller.extend({
   partialUndesignateBackLinkpath: Ember.computed.localStorage(),
 
   init() {
-    this.set('promptUserPopUp', false);
+    this.set("promptUserPopUp", false);
+  },
+
+  cannotDispatchItems(designation) {
```
namrataukirde: Is this criteria is same as dispatching from item details page? 

Just confirming :)

abhinav098: Yes

app/controllers/items/partial_undesignate.js
```diff
@@ -13,57 +15,88 @@ export default Ember.Controller.extend({
...
+    return (
+      designation &&
```
Darshan-Chauhan: can we set this up in a const and return that const?? As it would be more readable

abhinav098: yeah, that can be done.

PR #426, opened 2019-04-08
GCW-2309 fix css issues on stock

abhinav098:
Hi team please review this PR for some misalignment issues on stock app.
Thanks

### DIFF
--- a/app/styles/_search.scss
+++ b/app/styles/_search.scss
@@ -26,7 +26,7 @@
     @media #{$small-only} {
       font-size: 1.2rem;
       padding: 0.5rem 0.5rem;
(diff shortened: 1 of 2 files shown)


app/styles/_search.scss
```diff
@@ -91,8 +91,8 @@
   }
 
   .list.list-activity {
-    margin-left: 0rem;
-    margin-right: 0rem;
+    margin-left: 0.4rem;
```
abulsayyad123: `margin: 0 0.4rem` should work. No seperate left and right need to specify.

abhinav098: sure, just a sec, will change that.

PR #427, opened 2019-04-10
GCW-2429-User search keeps text, but the results are not correct

abulsayyad123:
Hi Team,
This PR is for clearing search text on changing page from search_users page during order creation. Please review the PR and let me know if any change is required.
Thanks

### DIFF
--- a/app/routes/order/search_users.js
+++ b/app/routes/order/search_users.js
@@ -34,8 +34,7 @@ export default AuthorizeRoute.extend({
     controller.set("filteredResults", users.uniq());
   },
 
-  deactivate() {
-    let controller = this.controllerFor("order.searchUsers");
+  resetController(controller) {
     controller.set("searchText", "");
   }
 });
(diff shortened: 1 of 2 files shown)


app/routes/order/search_users.js
```diff
@@ -1,31 +1,41 @@
...
         let organisationsUser = controller.findOrganisationsUser(org_user_id);
         this.store.pushPayload(organisationsUser);
       });
       users.push(userObject);
     });
 
     controller.set("filteredResults", users.uniq());
+  },
+
+  deactivate() {
```
patrixr: I think you can probably user `resetController` 

```
resetController(ctrl) {
  ctrl.set("searchText", "");
}
```

to avoid `this.controllerFor("order.searchUsers");`

abulsayyad123: changed `deactivate` with `resetController`.

abhinav098:
👍

PR #428, opened 2019-04-13
FIX => Wrong redesignate popup on item search page

abulsayyad123:
Hi Team,
This is fix for wrong popup for designated item on item search page. The issue was due to previous queryparams not getting cleaned up. Issue is fixed.
Thanks.

### DIFF
--- a/app/routes/items/search_order.js
+++ b/app/routes/items/search_order.js
@@ -84,5 +84,12 @@ export default AuthorizeRoute.extend({
     }
     controller.set("searchText", "");
     controller.set("backLinkPath", this.get("itemDesignateBackLinkPath"));
+  },
+
+  resetController(controller, isExiting) {
+    if (isExiting) {
+      controller.set("partial_qty", undefined);
+      controller.set("isSet", undefined);
+    }
   }
 });


app/routes/items/search_order.js
```diff
@@ -84,5 +84,12 @@ export default AuthorizeRoute.extend({
     }
     controller.set("searchText", "");
     controller.set("backLinkPath", this.get("itemDesignateBackLinkPath"));
+  },
+
+  resetController(controller, isExiting) {
+    if (isExiting) {
+      controller.set("partial_qty", undefined);
```
Darshan-Chauhan: why are we setting undefined?? I would very much prefer null

abulsayyad123: On doing `null` url does not remove these params and remains with value `Nan`. It removed it on assigning `undefined`.

patrixr: Ember object should really have an `unset` method. There's no clear way of deleting a key :/

patrixr: Just noticed this `controller.set("isSet", undefined);` 

--> could we make the variable name of `isSet` a bit more explicit ? I'm not really sure that is set or not 😆

abhinav098: right @patrixr, the `isSet` property is not very self-explanatory. @abulsayyad123 could you please make this something understandable ?

abhinav098: @patrixr , just got to know
set is collection of packages.. eg =>  Dining table is set which is made of 4 chairs and 1 table
you will find set_item_id column on package table maybe. 
isSet checks that the package is a set or not. 

cc - @namrataukirde @abulsayyad123

PR #429, opened 2019-04-15
Gcw 2535 add preferred contact number to summary tab

namrataukirde:
Hi Team,

This PR adds preferred contact number on the summary tab of the order.

Please review.

Thanks.

### DIFF
--- a/app/models/user.js
+++ b/app/models/user.js
@@ -1,46 +1,66 @@
-import attr from 'ember-data/attr';
-import Ember from 'ember';
-import { hasMany } from 'ember-data/relationships';
-import Addressable from './addressable';
+import attr from "ember-data/attr";
+import Ember from "ember";
+import { hasMany } from "ember-data/relationships";
+import Addressable from "./addressable";
 
 export default Addressable.extend({
-  firstName:   attr('string'),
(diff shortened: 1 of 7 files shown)


app/models/user.js
```diff
@@ -1,46 +1,66 @@
...
     const organisationsUsers = this.get("organisationsUsers");
-    return organisationsUsers.length ? 
-    organisationsUsers.get("firstObject.position") : '';
+    return organisationsUsers.length
+      ? organisationsUsers.get("firstObject.position")
+      : "";
+  }),
+
+  preferredContactNumber: Ember.computed("organisationsUsers.[]", function() {
+    let organisationsUser = this.get("organisationsUsers.firstObject");
```
abhinav098: Instead of `firstObject`, is there any other way? Maybe filter by `currentUserId` ?

namrataukirde: nope.  It should not work that way as logged in user will be staff member and we want preferred_contact_number of charity user.

created_by will be always for whom we are creating order. thus we will take that users preferred_ccontact_number which is stored in organisations_user record.

abhinav098: Okay makes sense. 👍

PR #432, opened 2019-04-17
Filtering orders by due date

patrixr:
## New

- Created a minimal `date-picker` component. The others were all custom made and had too much logic in it. Hopefully we can start moving towards pure generic configurable components like this

- `mock-utils.js` : A helper class with common mocks that we use everywhere. It removes a **lot** of the test setup code.

- A new filter page for orders. There are now 3 buttons on the search page :
(new button)
Screenshot 2019-04-17 at 4 31 23 PM

(new page)
Screenshot 2019-04-17 at 4 31 12 PM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ testem.log
 
 # cordova
 cordova/www
+cordova/assets/www
 cordova/platforms
 cordova/plugins/*
 cordova/.secret_keys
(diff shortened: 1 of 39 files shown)


patrixr:
I'll fast track this to Staging since Matt wants to test ASAP

PR #433, opened 2019-04-24
FIX => Filter array was being modified

patrixr:
Hi, 

A lot of this code is some cleanup and refactoring (because I have a problem and can't stop myself)

But the real bug comes from this line :

`filter.shift();`

This `shift` function **_mutates_** the array, which is never a good idea. We **always** want to write immutable code.  Here it (accidentally) modifies the data of another module, the best way to avoid those potential issues, is to never modify anything (leading to functional programming)

One other potential issue here, was that the code assumed that a certain element was placed first in an array. (assumptions in themselves are dangerous, we want to build deterministic systems)
So here, to remove a certain item from an array, I replaced a `.slice(1)`  with a `filter` call, which leaves no room for ambiguity

### DIFF
--- a/app/components/order-filters.js
+++ b/app/components/order-filters.js
@@ -1,5 +1,6 @@
 import Ember from "ember";
 import _ from "lodash";
+import { TYPE_FILTERS, STATE_FILTERS } from "../services/filter-service";
 
 // --- Helpers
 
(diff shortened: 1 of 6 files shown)


patrixr:
I'm going to fast track this for Matt to test. Please review regardless : )

PR #436, opened 2019-04-25
#April2019Release2: added missing translations

namrataukirde:
- Added missing translation for preferred contact number


namrataukirde:
Note: All chinese translations will be added in different PR.

PR #437, opened 2019-04-26
Filter buttons:  Clear filters

patrixr:
This PR adds a little cross next to each filter button, allowing us to clear them without opening the actual page : 

Screenshot 2019-04-26 at 12 15 25 PM


namrataukirde:
Great 👍

PR #448, opened 2019-05-09
GCW- 2577- updating cancelled Orders Packages

abulsayyad123:
Hi Team,
This PR is for JIRA ticket GCW-2577 (https://jira.crossroads.org.hk/browse/GCW-2577). I have added change for updating cancelled orders package in orders.

### DIFF
--- a/app/components/designate-form.js
+++ b/app/components/designate-form.js
@@ -240,40 +240,36 @@ export default Ember.Component.extend({
   },
 
   updatePartialQuantityOfSameDesignation() {
-    const order = this.get("order");
-    const url = this._buildDesignateUrl();
+    const item = this.get("item");
+    const url = `/items/${item.get(
+      "id"
+    )}/update_partial_quantity_of_same_designation`;
     const props = _.extend(this.itemDesignateParams(), {
       state: "cancelled",
       cancelled_orders_package_id: this.get("orderPackageId")
     });
-    this.showLoadingSpinner();
-    return this._ajaxRequest(url, "PUT", {
+
+    return this._ajaxUpdate(url, {
       package: props
-    })
-      .then(data => {
-        this.get("store").pushPayload(data);
-        this.get("router").transitionTo("orders.detail", order);
-      })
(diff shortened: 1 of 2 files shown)


app/components/designate-form.js
```diff
@@ -244,6 +244,20 @@ export default Ember.Component.extend({
         orders_package_id: ordersPackage ? ordersPackage.id : null //id of previous OrdersPackage ie to be cancelled
       };
 
+      var isSameDesignation = this.get("order.ordersPackages")
```
patrixr: I feel this function has grown too big 😄

app/components/designate-form.js
```diff
@@ -195,6 +195,44 @@ export default Ember.Component.extend({
     };
   }),
 
+  itemDesignateParams() {
```
patrixr: The chunk of code below is a bit mystical and unclear to me to be honest. Took me a significant amount of reads to understand 😄 

Few thoughts/questions

- where does this `cancelledState` come from ?

- You can probably refactor the url by code by doing `url = '/items/${id}/${action}'`

- Since the largest portion of the code is now in a huge param builder function, maybe creating methods which run the actions could be more understandable ? For clarity maybe using the API actions as function names could be good.
It would also allow us to add behaviours later on, as separate methods 

e.g
```javascript
isSameDesignation() { ... } // extract singular logics into methods

itemDesignateParams() { ... } // generic params

updatePartialQuantityOfDesignation() { ... } // action 1

designatePartialItem() { // action 2
   const url = this.buildUrl('designate_partial_item');
   const props = _.extend(this.itemDesignateParams(), { 
      ...
   });

  return ajax(...);
}

designateItem() { // main logic
  ...
  if (this.isSameDesignation()) {
     // We update the quantity
     return this.updatePartialQuantityOfDesignation();
  } else {
     // We re-designate to a new item
     return this.designatePartialItem();
  }
}
```

Those are just my subjective thoughts and opinions, I like to push for maintainable and easily upgradable code 😄 🚀

PR #449, opened 2019-05-10
GCW-2534- Organisation not searchable

abulsayyad123:
Hi Team,
This PR is for JIRA ticket GCW-2534- Organisation not searchable with half organisation name. Possible reason for not displaying the expected results is because we are restricting only 25 records from api. I have added pagination along with scrolling in the search organisation page. This will fix this issue.
Please review the PR and let me know if any changes is required.

### DIFF
--- a/app/controllers/search_organisation.js
+++ b/app/controllers/search_organisation.js
@@ -3,32 +3,21 @@ import searchModule from "./search_module";
 
 export default searchModule.extend({
   minSearchTextLength: 3,
+  displayResults: false,
 
-  onSearchTextChange: Ember.observer("searchText", function(){
-    if(this.get('searchText').length){
-      Ember.run.debounce(this, this.applyFilter, 500);
-    } else {
-      this.set("filteredResults", []);
+  onSearchTextChange: Ember.observer("searchText", function() {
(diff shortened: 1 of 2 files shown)


app/controllers/search_organisation.js
```diff
@@ -3,39 +3,39 @@ import searchModule from "./search_module";
 
 export default searchModule.extend({
```
patrixr: since you're using the new list, you can remove the searchModule inheritance.
It's also worth considering whether we want to make this generic in some way (there's quite a few that are exactly like this one)

patrixr:
👍

app/controllers/search_organisation.js
abulsayyad123: Yes, you are right. I thought to add some common code later in `searchModule` and refactor it later when all pages using searchModule have been modified to new list.

patrixr: It's probably better to let searchModule die, and re-create a new generic controller or component from scratch 😄 
that'll save you a lot of trouble

namrataukirde: Agree with @patrixr

PR #451, opened 2019-05-13
FIX=> For restricting multi quantity item search on Order Items Search

abulsayyad123:
Hi Team,
This PR is for restricting multi quantity item search on Order Items Search page. Please review the PR and let me know if any change is required.
Thanks

### DIFF
--- a/app/controllers/orders/items.js
+++ b/app/controllers/orders/items.js
@@ -33,7 +33,10 @@ export default searchModule.extend({
             return;
           }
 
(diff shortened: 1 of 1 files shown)


abhinav098:
👍

Darshan-Chauhan:
This should be fixed on API side. We shouldn't send data which is not needed

patrixr:
> This should be fixed on API side. We shouldn't send data which is not needed

I agree with that. Maybe adding a query option `?allowMultiQuantity=true|false`

PR #453, opened 2019-05-14
FIX: images from previous items were not being cleared

patrixr:
Hi team,

This is a first fix for the broken images. We know for a fact it'll fix some of the problems, but we don't know if it'll fix everything.
This whole page is pretty flawed, so it might still be possible to break it from here. The big question is whether another part of the app is also responsible for the broken images.

We also found cloudinary has a restore function, so I will write a script to try and restore as many of the images that were deleted as possible

FYI, I only added this line `controller.set("imageKeys", "");` , the rest is prettier


namrataukirde:
:+1:

PR #454, opened 2019-05-15
Release: Hotfix for broken images

patrixr:
As mentioned before, `controller.set("imageKeys", "")` is the only line added here.

@steveyken is also working on disabling the image deletion on the api until we get to the bottom of this.

Other fixes are likely to be added later on as well.


patrixr:
Cancelling the hotfix since we're going to make an actual release instead

PR #456, opened 2019-05-15
FIX broken process list due to headers not being sent

patrixr:
This is a temporary fix, we will work to move the fix in goodcity-lib afterwards.
We need to release ASAP as this is a blocker for the team


namrataukirde:
Great. Can be merged to staging and we can verify and release to live.

PR #468, opened 2019-05-30
GCW- 2421- Add total request count to order's contact summary

abulsayyad123:
Hi Team,
This PR is for ticket GCW-2421- Adding total request counts. Since, we have common routes for fetching `orders` data in nested routes. Which will results in  unnecessary request i.e `orders_count`, for data which is not required in other pages. So, I have added separate ajax request for `orders_count` on  `orders_summary` page instead of adding it in `users` or `orders` payload.

### DIFF
--- a/app/routes/orders/contact_summary.js
+++ b/app/routes/orders/contact_summary.js
@@ -1,10 +1,18 @@
-import detail from './detail';
+import detail from "./detail";
+import AjaxPromise from "stock/utils/ajax-promise"; //jshint ignore:line
 
 export default detail.extend({
-  setupController(controller, model){
-    if(controller) {
+  async setupController(controller, model) {
+    if (controller) {
       this._super(controller, model);
-      controller.set('isActiveSummary', true);
+      let userId = model.get("createdBy.id");
+      let ordersCount = await new AjaxPromise(
+        `/users/${userId}/orders_count`,
+        "GET",
+        this.session.get("authToken")
+      );
+      controller.set("ordersCount", ordersCount);
+      controller.set("isActiveSummary", true);
     }
   }
 });
--- a/app/routes/orders/contact_summary.js
+++ b/app/routes/orders/contact_summary.js
@@ -2,16 +2,24 @@ import detail from "./detail";
 import AjaxPromise from "stock/utils/ajax-promise"; //jshint ignore:line
 
 export default detail.extend({
-  async setupController(controller, model) {
-    if (controller) {
-      this._super(controller, model);
-      let userId = model.get("createdBy.id");
-      let ordersCount = await new AjaxPromise(
+  async model() {
+    const order = await this._super(...arguments);
+    const userId = order.get("createdBy.id");
+
+    return Ember.RSVP.hash({
+      order,
+      ordersCount: await new AjaxPromise(
         `/users/${userId}/orders_count`,
         "GET",
         this.session.get("authToken")
-      );
-      controller.set("ordersCount", ordersCount);
+      )
+    });
+  },
+
+  setupController(controller, model) {
+    if (controller) {
+      controller.set("model", model.order);
+      controller.set("ordersCount", model.ordersCount);
       controller.set("isActiveSummary", true);
     }
   }
--- a/app/routes/orders/contact_summary.js
+++ b/app/routes/orders/contact_summary.js
@@ -2,24 +2,22 @@ import detail from "./detail";
 import AjaxPromise from "stock/utils/ajax-promise"; //jshint ignore:line
 
 export default detail.extend({
+  orderService: Ember.inject.service(),
+
   async model() {
     const order = await this._super(...arguments);
     const userId = order.get("createdBy.id");
 
     return Ember.RSVP.hash({
       order,
-      ordersCount: await new AjaxPromise(
-        `/users/${userId}/orders_count`,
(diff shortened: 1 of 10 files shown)


app/routes/orders/contact_summary.js
```diff
@@ -1,10 +1,18 @@
...
 
 export default detail.extend({
-  setupController(controller, model){
-    if(controller) {
+  async setupController(controller, model) {
+    if (controller) {
       this._super(controller, model);
-      controller.set('isActiveSummary', true);
+      let userId = model.get("createdBy.id");
+      let ordersCount = await new AjaxPromise(
```
Darshan-Chauhan: `setupController` as name suggests should only be used to set data on controller. It shouldn't request data from API. Use `model` hook for that. Use nested promises if needed.

app/templates/orders/contact_summary.hbs
```diff
@@ -115,14 +115,37 @@
...
 
-      {{!-- <div class="row">
+      <div class="row">
         <div class="small-4 columns">
-          Total requests
+          {{t "order_contact_summary.total_requests"}}
         </div>
         <div class="small-8 columns email">
-          {{display-timeago model.createdBy.createdAt}}
+          {{#if ordersCount.submitted}}
```
Darshan-Chauhan: Can we loop through the states and use single if statement?

tests/acceptance/order-organisation-and-contact-test.js
```diff
@@ -85,6 +86,8 @@ module("Acceptance: Order summary", {
       }
     });
 
+    $.mockjax({ url: `/api/v1/users/*`, responseText: orders_count });
```
Darshan-Chauhan: Add this to a variable and clear after test case is finished running.

app/routes/orders/contact_summary.js
```diff
@@ -1,10 +1,26 @@
...
-    if(controller) {
-      this._super(controller, model);
-      controller.set('isActiveSummary', true);
+  async model() {
+    const order = await this._super(...arguments);
+    const userId = order.get("createdBy.id");
+
+    return Ember.RSVP.hash({
+      order,
+      ordersCount: await new AjaxPromise(
```
patrixr: We should move all those loose and wild api calls into services, we really shouldn't be doing that this way, it's not maintainable and it clutters the code.

--> not really a comment on this code specifically, just saw an opportunity to mention it. 
If anyone has an opportunity to take the lead in this change wants to create those services, you have my blessing 🤗 (also do drop me a message if you wish to get involved, I have a few tips)

cc - @abulsayyad123 @namrataukirde @Darshan-Chauhan @abhinav098 @steveyken

abulsayyad123: Yes.. that would be good thing to consider. Angular 2+ also, wraps all there ajax request request into singleton services.

steveyken: yes that seems like a very sensible idea. wrapping into a service provides a single place to maintain the code calls e.g. if the API changes or it's required elsewhere. Also simplifies testing :)

app/services/api-base-service.js
```diff
@@ -0,0 +1,70 @@
...
+    return this._request(
+      url,
+      {
+        action: "GET"
+      },
+      authorizedRequest
+    );
+  },
+
+  POST(url, opts = {}) {
```
patrixr: I would argue the payload (params) should be a separate argument from the opts :)

patrixr: Same goes for PUT

app/services/api-base-service.js
```diff
@@ -0,0 +1,70 @@
+import Ember from "ember";
```
patrixr: is it me or does it look like the whole file is indented 🤔 ?
Not that it matters much

app/services/order-service.js
```diff
@@ -0,0 +1,7 @@
+import ApiBaseService from "./api-base-service";
```
patrixr: I like this, implementing api calls become much simpler

patrixr:
> One main question - we are loading the count in via an ajax call. why not just include the request count in the user_profile payload when the app loads up?

@steveyken If I'm not mistaken, those are counts are not for the session's user. We open the page of an order, click on a tab, and only then do we see information of the person for whom it's submitted. That's when they get loaded.
We could consider having them in the user body, but that would probably mean it gets computed at times we don't really want them (even if we don't open that tab).
Just thinking out loud 😄

app/services/api-base-service.js
abulsayyad123: It's properly indented :)

steveyken:
Great that makes sense. I hadn't read the spec properly.

app/services/api-base-service.js
patrixr: ok, github was just trying to trick me : )

app/services/api-base-service.js
abulsayyad123: I just thought to reduce number of arguments.

app/services/api-base-service.js
```diff
@@ -0,0 +1,69 @@
...
+import AjaxPromise from "./../utils/ajax-promise";
+
+export default Ember.Service.extend({
```
abhinav098: Looking great... 👍

PR #472, opened 2019-06-10
Refactor inline-select-list

patrixr:
## Changes

- Removed hardcoded lookup lists for donorConditions
- Removed hardcoded lookup lists for identityTypes
- Refactor of inline-select-list to be generic
- Deletion of beneficiary-inline-select-list
- Upon logout, we don't `unloadAll()` as it also removes preloaded lookup tables. We unload sensitive data explicitely with `session.unloadSessionData()` --> Haven't found a nicer way to do this, open to suggestions !

### Generic components

Components should be **generic**, meaning they should **never** contain model-specific code, or page-specific code because that prevents them from being **reusable**.
We're very far from this idea currently, but we need to work towards that.

You can see here that the `inline-select-list` component had conditions such as 
```javascript
if(name === "donor_condition_id")  {
   ...
}
```

We can actually see that in action here --> the list component was so hard to re-use that a `beneficiary-inline-select-list` had been created for another page. An approach which is unmaintainable. 

I've deleted it and made the base component generic for it to work in both pages.
If you find code like this within a ui component, time to refactor 😎 

### Business logic within components

Less dramatic but still not great, a component making API calls is also something we want to avoid. They are designed to be pieces of the UI, which we can use everywhere. No business logic should be found there.

In this case, to make the transition easier, I've made saving the model turned off by default, and it needs to be enabled explicitely with `autosave=true`

Fun fact ( 😄), In component based frameworks (react, vue), since everything is a component, people tend to create folders like `views/` and `components`, respectively containing main app components from generic components.


abhinav098:
PR looks good @patrixr but there are some code climate issues. Could you please take a look at those ?

patrixr:
> PR looks good @patrixr but there are some code climate issues. Could you please take a look at those ?

@abhinav098 prettier at work again 😄
I refactored `undispatchFullQuantity()` , but left the other ones since they are in the search_module which we're phasing out.

namrataukirde:
Excellent. 👌 :)

PR #474, opened 2019-06-12
GCW-2455- Add show more button for long Purpose Description

abulsayyad123:
Hi Team,
This PR has feature of showing `show more` button for long purpose description inside client summary tab. Please review the PR and let me know if any change is required.
Thanks

### DIFF
--- a/app/components/order-inline-textarea-input.js
+++ b/app/components/order-inline-textarea-input.js
@@ -60,7 +60,7 @@ export default AutoResizableTextarea.extend({
         .show();
       return false;
     }
-    this.get("enableShowMore")(false);
+    this.get("enableDescriptionEdit")({ makeEditable: false });
     Ember.$(this.element).removeClass("item-description-textarea");
   },
 
--- a/app/components/order-inline-textarea-input.js
+++ b/app/components/order-inline-textarea-input.js
@@ -23,13 +23,12 @@ export default AutoResizableTextarea.extend({
   },
 
   focusOut() {
-    var order = this.get("order");
-    var url = `/orders/${order.get("id")}`;
-    var key = this.get("name");
-    var value = this.attrs.value.value || "";
-    var orderParams = {};
+    const orderId = this.get("order.id");
+    const key = this.get("name");
+    const value = this.attrs.value.value || "";
+    let orderParams = {};
     orderParams[key] = this.get("value").trim() || "";
-    var element = this.element;
+    let element = this.element;
 
     if (
       orderParams[key].toString() !==
@@ -38,12 +37,10 @@ export default AutoResizableTextarea.extend({
           .trim() &&
       value !== ""
     ) {
-      var loadingView = getOwner(this)
+      let loadingView = getOwner(this)
         .lookup("component:loading")
         .append();
-      new AjaxPromise(url, "PUT", this.get("session.authToken"), {
-        order: orderParams
-      })
+      this.get("updatePurposeDescription")(orderId, { order: orderParams })
         .then(data => {
           this.get("store").pushPayload(data);
           Ember.$(".client_summary_description_error").hide();
--- a/app/components/order-inline-textarea-input.js
+++ b/app/components/order-inline-textarea-input.js
@@ -46,9 +46,7 @@ export default AutoResizableTextarea.extend({
       })
         .then(data => {
           this.get("store").pushPayload(data);
-          Ember.$(element)
-            .siblings()
-            .hide();
+          Ember.$(".client_summary_description_error").hide();
         })
         .finally(() => {
           loadingView.destroy();
--- a/app/components/order-inline-textarea-input.js
+++ b/app/components/order-inline-textarea-input.js
@@ -1,52 +1,69 @@
-import Ember from 'ember';
-import AutoResizableTextarea from './auto-resize-textarea';
-import AjaxPromise from 'stock/utils/ajax-promise';
+import Ember from "ember";
+import AutoResizableTextarea from "./auto-resize-textarea";
+import AjaxPromise from "stock/utils/ajax-promise";
 const { getOwner } = Ember;
 
 export default AutoResizableTextarea.extend({
-  previousValue: '',
+  previousValue: "",
   store: Ember.inject.service(),
   order: null,
-  orderCopy: Ember.computed.readOnly('order'),
+  orderCopy: Ember.computed.readOnly("order"),
 
   keyDown() {
     var value = this.element.value;
-    if( value.charCodeAt(value.length - 1) === 10 && event.which === 13) {
+    if (value.charCodeAt(value.length - 1) === 10 && event.which === 13) {
       return false;
     }
   },
 
   didInsertElement() {
-    Ember.$('.description-textarea').val(this.get('orderCopy.data.purposeDescription'));
+    Ember.$(".description-textarea").val(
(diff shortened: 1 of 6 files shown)


app/components/order-inline-textarea-input.js
```diff
@@ -1,60 +1,75 @@
...
+      orderParams[key].toString() !==
+        this.get("previousValue")
+          .toString()
+          .trim() &&
+      value !== ""
+    ) {
+      var loadingView = getOwner(this)
+        .lookup("component:loading")
+        .append();
+      new AjaxPromise(url, "PUT", this.get("session.authToken"), {
```
patrixr: I understand that it's hard to change, but that's a good example of a visual component with business logic. Which makes it very hard to re-use.

app/controllers/orders/client_summary.js
```diff
@@ -17,29 +21,78 @@ export default detail.extend({
...
     ];
   }),
 
+  isLongDescription: Ember.computed("model.purposeDescription", function() {
+    return this.get("model.purposeDescription.length") > LONG_DESCRIPTION_SIZE;
+  }),
+
+  toggleShowMore: Ember.observer("isLongDescription", function() {
+    const isLongDescription = this.get("isLongDescription");
+    if (isLongDescription) {
```
patrixr: you could simply do `this.set("showMore", isLongDescription)` 

note: I don't believe that there's actually anything wrong with the if/else here, it's still pretty understandable and readable to me.

app/controllers/orders/client_summary.js
```diff
@@ -17,29 +21,78 @@ export default detail.extend({
...
+      this.set("showMore", false);
+    }
+  }),
+
+  shortPurposeDescription: Ember.computed(
+    "showInlineEdit",
+    "showMore",
+    function() {
+      let purposeDescription = this.get("model.purposeDescription");
+      if (this.get("showMore")) {
```
patrixr: shouldn't it be the opposite ? "show**More**"  should be be showing more, but here's it's cutting the text 🤔

app/controllers/orders/client_summary.js
abulsayyad123: In case when description is one liner, we can't show `show more` . that is why I am explicitly hiding it.

patrixr: I know, I just meant you could shorten that piece of code : )

app/controllers/orders/client_summary.js
```diff
@@ -17,29 +22,80 @@ export default detail.extend({
 
   identityTypes: Ember.computed(function() {
```
patrixr: Those should not be hardcoded, identityTypes are a lookup table

app/controllers/orders/client_summary.js
```diff
@@ -17,29 +22,80 @@ export default detail.extend({
...
+    showMoreToggle() {
+      this.toggleProperty("displayShowMore");
+    },
+
+    enableDescriptionEdit(opts = {}) {
+      const { makeEditable = true } = opts;
+      this.set("showInlineEdit", makeEditable);
+      if (makeEditable) {
+        Ember.run.later(() => {
+          Ember.$("#desc").click();
```
patrixr: what is this for ? 🤔

app/controllers/orders/client_summary.js
```diff
@@ -17,29 +22,80 @@ export default detail.extend({
...
+    updatePurposeDescription(orderId, body) {
+      return this.get("orderService").updateOrder(orderId, body);
+    },
+
     removeBeneficiaryModal() {
-      this.toggleProperty('showBeneficiaryModal');
+      this.toggleProperty("showBeneficiaryModal");
     },
 
     deleteBeneficiary() {
```
patrixr: this has been refactored in master, make sure to keep the fresh version : )

https://github.com/crossroads/stock.goodcity/blob/403baf2837fef8a1476eee4abc721103ddba0a19/app/controllers/orders/client_summary.js#L27

app/controllers/orders/client_summary.js
```diff
@@ -17,29 +22,80 @@ export default detail.extend({
...
+      const { makeEditable = true } = opts;
+      this.set("showInlineEdit", makeEditable);
+      if (makeEditable) {
+        Ember.run.later(() => {
+          Ember.$("#desc").click();
+        }, 50);
+      }
+    },
+
+    updatePurposeDescription(orderId, body) {
```
patrixr: If this is named `updatePurposeDescription`, then it should probably take a "description" argument instead of "body"

Note that you could also do
```javascript
order.set('purposeDescription', description)
order.save()
```

app/components/order-inline-textarea-input.js
```diff
@@ -1,60 +1,72 @@
-import Ember from 'ember';
```
patrixr: Although it's a great start having it take a callback as argument and removing the API call. It's still only usable for orders.

You could just have a generic `inline-textarea` component that sends the updated text when it changes : )
Building the `orderParams` is not the component's responsibility.

Imagine I want the exact same textinput for another property of another model, we don't want to be forced to re-write the whole component.

Analogy: the component is a car, and the controller is the driver. The car doesn't decide which route to take, it's just a tool which the driver can use to achieve his/her goal. It has a wheel, pedals, and you can use it in different ways.

app/controllers/orders/client_summary.js
abulsayyad123: This is for making textarea editable by clicking.

patrixr: but, shouldn't just clicking on it naturally set focus to it ? 🤔

app/components/order-inline-textarea-input.js
abulsayyad123: That make sense. I have renamed it and also pulled out Business logic from the component.

app/controllers/orders/client_summary.js
```diff
@@ -19,7 +23,60 @@ export default detail.extend({
...
+    isLongDescription
+      ? this.set("displayShowMore", true)
+      : this.set("displayShowMore", false);
+  }),
+
+  shortPurposeDescription: Ember.computed(
+    "showInlineEdit",
+    "displayShowMore",
+    function() {
+      let purposeDescription = this.get("model.purposeDescription");
```
patrixr: if one day you need to re-use this logic, it'd be cool to have it as a helper

e.g `{{ellipsis-text maxChar=170 collapsed=displayShowMore text=purposeDescription }}`

Not a change request, just a thought : )

app/controllers/orders/client_summary.js
```diff
@@ -19,7 +23,60 @@ export default detail.extend({
...
+    },
+
+    updatePurposeDescription(purposeDescription) {
+      const order = this.get("model");
+      this.showLoadingSpinner();
+      order.set("purposeDescription", purposeDescription);
+      order
+        .save()
+        .then(() => {
+          Ember.$(".client_summary_description_error").hide();
```
patrixr: we want to avoid using jquery as a general rule of thumb, and rather use a property with and `#if` hbs condition

`set('showError', true)`

Same goes for setting/unsetting classes, examples available [here](https://guides.emberjs.com/release/templates/conditionals/)

abulsayyad123:
Closing this PR. Since, we have to refactor code of the editable input boxes. This PR is no longer useful. But we can refer its logic for future reference.

PR #476, opened 2019-06-17
redirect users to login page instead of showing modal

abhinav098:
Currrently, if user is logged out and tries to visit any page, we show "You must login!" modal, which looks rude, so this PR removes that modal, and automatically redirects them to the login page.

Keeping in mind which the user was trying to visit, post login user is redirected to that page.

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -11,7 +11,6 @@ export default Ember.Route.extend(preloadDataMixin, {
   logger: Ember.inject.service(),
   messageBox: Ember.inject.service(),
   cordova: Ember.inject.service(),
-  isMustLoginAlreadyShown: false,
 
   _loadDataStore: function() {
     return this.preloadData()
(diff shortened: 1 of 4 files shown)


app/routes/application.js
```diff
@@ -1,84 +1,97 @@
...
+      var authToken = window.localStorage.getItem("authToken");
+      if (
+        !authToken &&
+        !(
+          currentPath.indexOf("login") >= 0 ||
+          currentPath.indexOf("authenticate") >= 0
+        )
+      ) {
+        object.session.clear();
+        object.store.unloadAll();
```
patrixr: there's a method in the `session` service called `unloadSessionData()`
It removes orders/items and such, but keeps lookup tables and static data. Should be more appropriate than deleting all the data : )

app/routes/application.js
```diff
@@ -1,84 +1,97 @@
...
+      ) {
         object.transitionTo("/");
       }
     };
-    window.addEventListener("storage", function() {
-      storageHandler(_this);
-    }, false);
+    window.addEventListener(
+      "storage",
+      function() {
```
patrixr: if you replace `storageHandler` with a lambda (`var storageHandler =(object) => { ... }`) there won't be a need for `_this` anymore

And you can do `window.addEventListener('storage', storageHandler, false)`

Not a big deal, just thought it's worth mentionning

app/routes/application.js
```diff
@@ -1,84 +1,97 @@
...
-        this.handleError(error);
-      }
-    }).finally(() => {
-      // don't know why but placing this before preloadData on iPhone 6 causes register_device request to fail with status 0
-      if (this.session.get('isLoggedIn')) {
-        this.get("cordova").appLoad();
-      }
-    });
+  _loadDataStore: function() {
+    return this.preloadData()
```
patrixr: should we maybe increase the line length of prettier ? this is horrible 😄

app/routes/application.js
abhinav098: sure, @patrixr , latest master was not on my local, now i have rebased it. 
Thanks for mentioning this.

app/routes/application.js
abhinav098: yeah, i think its configurable.

PR #480, opened 2019-06-20
Preload after login

patrixr:
Hi, this PR fixes an issue that @namrataukirde found yesterday.

That being: The preload routine of Stock does not run when a user logs in, it is only run if the user was previously logged in before and restores the session.

The reason this hasn't exploded before, is because the records that need to be preloaded usually arrive as dependencies of other requests. Also some pages load the data themselves in routes.

Changes:
- Run the preload after login
- Some code cleanup of things I stumbled upon

### DIFF
--- a/app/services/auth-service.js
+++ b/app/services/auth-service.js
@@ -1,0 +1,22 @@
+import ApiBaseService from "./api-base-service";
+
+export default ApiBaseService.extend({
+  opts: {
+    authorizedRequest: false
+  },
+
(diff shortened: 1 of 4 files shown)


app/services/auth-service.js
```diff
@@ -0,0 +1,22 @@
+import ApiBaseService from "./api-base-service";
+
+export default ApiBaseService.extend({
```
namrataukirde: Nice 👌

PR #493, opened 2019-08-23
GCW-2763 handle errors graciously

abhinav098:
Render error messages coming from the API.

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -65,7 +65,7 @@ export default Ember.Route.extend(preloadDataMixin, {
   },
 
   getErrorMessage(reason) {
-    if (reason.errors[0].detail && reason.errors[0].detail.status == 422) {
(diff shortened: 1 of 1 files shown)


app/routes/application.js
```diff
@@ -64,16 +64,20 @@ export default Ember.Route.extend(preloadDataMixin, {
     );
   },
 
-  showSomethingWentWrong(reason) {
+  getErrorMessage(reason) {
+    if (reason.errors[0].detail && reason.errors[0].detail.status == 422) {
```
Darshan-Chauhan:
```suggestion
    if (reason.errors.length && reason.errors[0].detail && reason.errors[0].detail.status == 422) {
```

patrixr: Another options :

```javascript
if (_.get(reason, 'errors[0].detail.status') == 422) {
```

PR #496, opened 2019-09-05
Gcw 2752 add preferred contact number field on user form

akshaygwd:
Hi Team,

This PR contains Addition of a preferred contact number field in the stock app for new user form,
It also contains the UI update for the same new user form.

Please review.
Thanks

### DIFF
--- a/app/controllers/users/new.js
+++ b/app/controllers/users/new.js
@@ -31,7 +31,7 @@ export default Ember.Controller.extend({
         .lookup("component:loading")
         .append();
       const mobilePhone =
-        this.get("preferredPhone.length") &&
+        this.get("mobilePhone.length") &&
         `${config.APP.HK_COUNTRY_CODE}${this.get("mobilePhone")}`;
       const preferredPhone =
         this.get("preferredPhone.length") && this.get("preferredPhone");
--- a/app/controllers/users/new.js
+++ b/app/controllers/users/new.js
@@ -30,11 +30,12 @@ export default Ember.Controller.extend({
       const loadingView = getOwner(this)
         .lookup("component:loading")
         .append();
+      const preferredContactvalue = this.get("preferredPhone");
       const mobilePhone =
         this.get("mobilePhone.length") &&
         `${config.APP.HK_COUNTRY_CODE}${this.get("mobilePhone")}`;
       const preferredPhone =
-        this.get("preferredPhone.length") && this.get("preferredPhone");
+        preferredContactvalue.length && preferredContactvalue;
       const firstName = this.get("firstName");
       const lastName = this.get("lastName");
       const organisationId = this.get("organisationId");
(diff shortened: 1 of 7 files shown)


app/controllers/users/new.js
```diff
@@ -22,55 +22,83 @@ export default Ember.Controller.extend({
...
     this.set("position", "");
+    this.set("preferredPhone", "");
   },
 
-  formatMobileNumber(){
-    var mobile = this.get('mobilePhone');
-    if(mobile.length){
-      return config.APP.HK_COUNTRY_CODE + this.get('mobilePhone');
+  formatMobileNumber(mobileNumber) {
+    if (mobileNumber.length) {
```
Darshan-Chauhan: I dont understand use case of this function.
We're passing a number and returning the same number in if and else both conditions. Do we even require this function?

app/controllers/users/new.js
```diff
@@ -22,55 +22,83 @@ export default Ember.Controller.extend({
...
   },
 
   actions: {
     saveUser() {
-      var loadingView = getOwner(this).lookup('component:loading').append();
-      var mobilePhone = this.formatMobileNumber();
-      var firstName = this.get('firstName');
-      var lastName = this.get('lastName');
-      var organisationId = this.get('organisationId');
+      var loadingView = getOwner(this)
```
Darshan-Chauhan: I think all `var` declarations can be replaced with `const.

akshaygwd: @Darshan-Chauhan Please review the PR I did the suggested changes

app/controllers/users/new.js
```diff
@@ -25,28 +25,21 @@ export default Ember.Controller.extend({
...
-      var lastName = this.get("lastName");
-      var organisationId = this.get("organisationId");
-      var position = this.get("position");
-      var email = this.get("email");
+      const mobilePhone =
+        this.get("preferredPhone.length") &&
+        `${config.APP.HK_COUNTRY_CODE}${this.get("mobilePhone")}`;
+      const preferredPhone =
+        this.get("preferredPhone.length") && this.get("preferredPhone");
+      const firstName = this.get("firstName");
```
Darshan-Chauhan: this.get('preferredPhone') can be cached

app/controllers/users/new.js
```diff
@@ -25,28 +25,21 @@ export default Ember.Controller.extend({
...
-        config.APP.HK_COUNTRY_CODE +
-        this.formatMobileNumber(this.get("mobilePhone"));
-      var preferredPhone = this.formatMobileNumber(this.get("preferredPhone"));
-      var firstName = this.get("firstName");
-      var lastName = this.get("lastName");
-      var organisationId = this.get("organisationId");
-      var position = this.get("position");
-      var email = this.get("email");
+      const mobilePhone =
+        this.get("preferredPhone.length") &&
```
Darshan-Chauhan: I think this should be this.get('mobilePhone.length').

PR #497, opened 2019-09-10
GCW-2773- Implement new item creation

abulsayyad123:
Hi Team,
This PR is for adding new fields and some UI modeification in `Add Item` screen.  
Ticket URL: https://jira.crossroads.org.hk/browse/GCW-2773 

The UI changes is similar to ticket : https://jira.crossroads.org.hk/browse/GCW-2697 (screens available on this ticket).

Changes: Along with new UI changes, i've done refactoring of `items/new.js` controller. Also separated `ajax` requests to separate services.

Thanks.


app/controllers/items/new.js
```diff
@@ -107,6 +120,36 @@ export default Ember.Controller.extend({
...
+    const height = this.get("height");
+    const dimensionsCount = _.filter(
+      [length, width, height],
+      item => Number(item) <= 0
+    ).length;
+    return _.inRange(dimensionsCount, 1, 3);
+  }),
+
+  printLabelCount: Ember.computed("labels", function() {
+    return (this.get("labels") * 1).toString();
```
swatijadhav: This should be sufficient
```suggestion
    return this.get("labels").toString();
```

Darshan-Chauhan: This change was for following reason.

We wanted to remove the initial 0's from labels value and display.

For example
```
this.get('labels') gives `"00999"`

So we want to display `"999"` only

Hence `("00999" * 1).toString()` gives `"999"`
```

patrixr: Maybe add a comment, there's a change someone might remove it otherwise 😄 

Other potential options:
- `Number(this.get('labels'))`
- `this.get('labels').replace(/^0+/, '')`

app/controllers/items/new.js
```diff
@@ -195,23 +240,25 @@ export default Ember.Controller.extend({
...
     this.set("locationId", "");
     this.set("inventoryNumber", "");
-    loadingView.destroy();
-    _this.replaceRoute("items.detail", data.item.id);
+    this.hideLoadingSpinner();
+    this.replaceRoute("items.detail", data.item.id);
   },
 
-  getItemConditions(_this) {
+  getItemConditions() {
```
patrixr: Feels like this should be named something like `hasIncompleteConditions` or `isInvalid`
`getXXX` is typically used for getters so a bit confusing : )

app/services/package-service.js
```diff
@@ -0,0 +1,19 @@
+import ApiBaseService from "./api-base-service";
+
+export default ApiBaseService.extend({
```
patrixr: clean ! 👌

app/controllers/items/new.js
Darshan-Chauhan: `Number(this.get('labels'))` looks much more readable and clean. 👍

app/controllers/items/new.js
patrixr: Would be nice to create a small validation util for models, e.g

```javascript
{
    hasValidConditions: Rule.create([
         { prop: 'description', rules: ['not_empty] },
         { prop: 'code', rules: ['exists'] },
         { prop: 'width', rules: [ ['gt', 0] ] } 
    ]) // => returns an ember computed property
}
```

And use that across the app

abulsayyad123: Yeah.. I have added this name `isInValidConditions` where it has been called. But yes, it would be better to rename it here too.

PR #498, opened 2019-09-12
#September2019Release1

abulsayyad123:
Hi Team,
This PR contains:
- District selection in logistic tab is not in alphabetical order (GCW-2508)
- Able to do `Start Dispatching` for the order even if the checklist is not checked for previously processed orders (GCW-2441)
- vulnerabilities fix.
- Handle Errors and render actual errors that api sends (GCW-2763)

Please review.


app/controllers/orders/detail.js
```diff
@@ -236,19 +236,24 @@ export default GoodcityController.extend({
...
-            this.send("changeOrderState", order, actionName);
-          } else {
-            this.genericAlertPopUp(
-              "order_details.logistics.checklist_incomplete"
-            );
-          }
+          this.send("verifyChecklistAndChangeState", order, actionName);
+          break;
+        case "start_dispatching":
+          this.send("verifyChecklistAndChangeState", order, actionName);
```
swatijadhav: case `finish_processing` and `start_dispatching` have same blocks, so they can be combined here.

PR #504, opened 2019-09-24
GCW-2807-Add Infinite scroll on Items filters and search Item for Orders page.

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2807

### What does this PR do?

FEATURE: Added infinite-scroll on Items and Orders item search page. Created `search` mixin to clean repeated codes.

Please review this PR.

### DIFF
--- a/app/controllers/items/index.js
+++ b/app/controllers/items/index.js
@@ -17,6 +17,10 @@ export default Ember.Controller.extend(SearchMixin, {
    * @type {Boolean}, expected in SearchMixin
    **/
   autoLoad: true,
+  /**
+   * @type {Number}, perPage in response
+   **/
+  perPage: 25,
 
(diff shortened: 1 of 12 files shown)


app/controllers/items/index.js
```diff
@@ -1,75 +1,70 @@
 import config from "../../config/environment";
-import Ember from "ember";
-import searchModule from "../search_module";
+import _ from "lodash";
+import SearchMixin from "../../mixins/search_resource";
```
swatijadhav: Can we use the absolute path instead of a relative path? I am also not sure, but @abulsayyad123 would you like to give a try?

```suggestion
import SearchMixin from "stock/mixins/search_resource";
```

abulsayyad123: It's working. Added absolute path suggestion.

PR #513, opened 2019-10-03
Fix: disable publish button if qty==0

patrixr:
I forgot to include this rule in the previous PR -- cc @namrataukirde

### DIFF
--- a/app/controllers/items/detail.js
+++ b/app/controllers/items/detail.js
@@ -32,6 +32,16 @@ export default GoodcityController.extend(singletonItemDispatchToGcOrder, {
     }
   ),
 
+  allowPublish: Ember.computed(
+    "model.isSingletonItem",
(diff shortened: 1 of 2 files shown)


app/controllers/items/detail.js
```diff
@@ -32,6 +32,16 @@ export default GoodcityController.extend(singletonItemDispatchToGcOrder, {
     }
   ),
 
+  allowPublish: Ember.computed(
+    "model.isSingletonItem",
```
Darshan-Chauhan: `Ember.computed('model.{isSingletonItem, availableQty}')`

patrixr: oh that's cool :)

patrixr: I'll merge as is, I'll tweak in my upcoming batch of changes 👍

PR #517, opened 2019-10-10
Item Details > OrdersPackages blocks

patrixr:
# OrdersPackages blocks

This ticket adds the orders_packages blocks to the item details page.

It is dependant on this [API CHANGE](https://github.com/crossroads/api.goodcity/pull/832), please read it first.

They look a little something like that : 

Screenshot 2019-10-10 at 3 34 41 PM

#### Side drawer

A **generic** component that slide _any_ content from the right
```handlebars
{{#side-drawer}}
    <div> This text will appear from the right when arrow is clicked </div>
{{/side-drawer}}
```

#### Action drawer

A **generic** special usage of the `side-drawer` which receives a list of actions and displays them in a drawer
```handlebars
{{action-drawer actionList=actionList}}
```

#### Orders Package Block

A **custom** component to represent a block, it contains an `action-drawer` populated by the actions the [API](https://github.com/crossroads/api.goodcity/pull/832) provided.

Note: I've placed this component under a new `goodcity/` folder.
The reason for that is to differentiate generic UI components (such as the drawers) from the custom logical

```handlebars
{{goodcity/orders-package-block orderPkg=orderPkg }}
```

#### Popup overlay

A **generic** component that slides up from the bottom of the screens to display content

```handlebars
{{#popup-overlay open=open}}
    <div> this will appear from the bottom of the screen when 'open' is true</div>
{{/popup-overlay}}
```

#### Alias (utility only)

A small component used to rename variables within the template file

```handlebars
 {{#let-alias model.someReallyLongPropertyName as |short| }}
     <div> This is {{short}} </div>
     <div> I can use {{short}} many times {{short}} </div>
     <div> because it's very {{short}} </div
 {{/let-alias}}
```

#### Toggle helper

Allows to create an action that toggles a property without needing to write it in javascript

```handlebars
 <button {{action (toggle this "myBoolProperty")}}>
   Value: {{ myBoolProperty }}
 </button>
```

### Designation Service

A new service class called `DesignationService` was added.
Notes:

> Stock currently use the 'designation' model, which in reality is the 'order' model.
>
> We're trying to move **away** from this model name.
>  '**Designation**' here refers to the assignment of a package to an order
> which mostly involves operations on the orders_package join table
> 
> 

This service contains one main method :  `execAction(ordersPackage, actionName, params)` which triggers the actions defined in the [API](https://github.com/crossroads/api.goodcity/pull/832)

### Service-backed component: Orders Search Overlay (**important**)

A **global component**, linked to the DesignationService, is now available.

When called upon, it will pop-up from under the screen (built using the `popup-overlay`) and will allow the user to search and select an order.

**> why not use the search_orders page** : 

- The `search_orders` page is overly complicated
- It is slow due to navigation time
- It has navigation/back issues (common everywhere)
- It triggers an api call (designation), instead of returning an order, which is not what its name suggests
- We can't animate the way the mockups do because it's a page

**> How to use the new service backed component**

**Anywhere**  in the code, simply do this : 

```javascript
const order = await designationService.userPickOrder( ... some filters if needed ...)

// Do something with order. Order is null is the user cancelled the search
```

This will open the popup, wait until the user selects an order, and return order.

After which you can just continue your function.

## Disabled features

- `Edit Quantity` is currently disabled and yet to be implemented
- `Redesignate` is currently disabled as a side-effect of the quantity being set to `0` when cancelling an orders_package (will surely change with the work Matt is doing on quantities)

## Ticket Link: 

Second part of https://jira.crossroads.org.hk/browse/GCW-2778

## Impacted Areas

Item Details > Publishing Tab

## Mockup Link

https://xd.adobe.com/view/224f7314-f1c0-40b6-7d7a-929e616be6bb-547e/screen/889371e6-fc7b-4556-9a0e-f4649361e6e9/iPhone-6-7-8-122/


app/components/goodcity/orders-package-block.js
```diff
@@ -0,0 +1,204 @@
...
+
+  hideSpinner() {
+    Ember.run(() => {
+      if (this.loadingView) {
+        this.loadingView.destroy();
+        this.loadingView = null;
+      }
+    });
+  },
+
```
abulsayyad123: Wouldn't it be better if we make common mixins which will have these uitlity for spinners, catching errors etc. Like similar we have in admin. We should import these mixins in root component, routes and controllers. Through which all components, controllers and routes should inherit.
Thoughts ?

app/templates/application.hbs
```diff
@@ -14,3 +14,12 @@
 {{/init-foundation}}
 {{outlet}}
 <div class="printer_message_block"></div>
+
+{{!-- Service-backed components --}}
+{{#let-alias designationService as |s|}}
+  {{goodcity/orders-search-overlay
+    open=s.openOrderSearch
+    onSelect=s.onnOrderSelected
```
steveyken: `onn` typo?

app/templates/components/let-alias.hbs
```diff
@@ -0,0 +1 @@
+{{yield value}}
```
steveyken: perhaps one of our most simple yet powerful templates in the entire project!

app/components/goodcity/orders-package-block.js
patrixr: Yeah, we have one in Browse (I think), but I got lazy 😄 
I'll bring it to Stock

patrixr: oh Admin, not browse. Like you said

app/templates/application.hbs
patrixr: Darn, my brokenn Mac keyboard keeps doubling the Ns and the Ts

PR #525, opened 2019-10-24
Undispatch action: Location overlay for selection

patrixr:
### Additions : 

- Location selection overlay (similar to the orders one)
- LocationService to control the overlay. Running `this.get("locationService").userPickLocation()` will trigger the overlay, and return when the user has selected a location

### Feature change : 

- When clicking on the `orders_package` undispatch action, the location overlay pops up to select the location


app/components/goodcity/orders-package-block.js
```diff
@@ -147,6 +163,9 @@ export default Ember.Component.extend(AsyncMixin, {
       await this.runTask(
         this.get("actionRunner").execAction(ordersPkg, actionName, params)
       );
+
+      // fire & forget
+      this.get("store").findRecord("item", itemId, { reload: true });
```
Darshan-Chauhan: Is this for updating an item?
If yes, are we not sending live updates for `orders_packages`? 🤔

app/models/item.js
```diff
@@ -37,13 +37,26 @@ export default cloudinaryUrl.extend({
...
   designateFullSet: Ember.computed.localStorage(),
 
+  isDesignated: Ember.computed(
+    "ordersPackages",
+    "ordersPackages.[]",
+    "ordersPackages.@each.state",
+    function() {
+      return (
+        this.get("ordersPackages")
+          .rejectBy("state", "requested")
```
Darshan-Chauhan: Will it be better to use filterBy?

```
this.get("ordersPackages").filterBy('state', 'designated').get("length") > 0
```

https://api.emberjs.com/ember/2.15/classes/Ember.Array/methods/filterBy?anchor=filterBy

app/components/goodcity/orders-package-block.js
patrixr: Yeah Steve and I found a weird edge case. Live updates are ignored when they are triggered by the session itself (for good reasons, apps would break because live updates would land before the requests finishes).

So it was either a background reload, or we fiddle/hack the subscriptions to let only some updates through. None of which is really great, we went for the lesser of two evils

app/models/item.js
patrixr: Yeah, it's actually the name that's wrong here. meant to behave as `isDesignatedOrDispatched`, I'll update it

patrixr: Actually I'll keep a proper `isDesignated` and make `isUnavailable`, makes more sense

patrixr: @Darshan-Chauhan updated 👍

app/components/goodcity/orders-package-block.js
patrixr: BTW, this is for a deletion rather than an update. The API has no way of letting the front-end know that a record was deleted, apart from live updates or reloading.

Darshan-Chauhan: I think `unloadRecord` would be a better choice here.
https://api.emberjs.com/ember-data/3.10/classes/DS.Model/methods/unloadRecord?anchor=unloadRecord

patrixr: We don't want to unload anything per-say. We just want the record in the same state as the backend, one of it's relationships may or may not have been deleted.
The JSON we receive from the backend is a snapshopt of the relationshipts, but ember just adds what it receives to the store, without doing a diff on what currently exists.
So we either refresh the record, or we let the live-updates through.

app/components/image-zoom.js
```diff
@@ -1,44 +1,53 @@
-import Ember from 'ember';
-import config from '../config/environment';
+import Ember from "ember";
+import config from "../config/environment";
```
swatijadhav: Use an absolute path.

PR #527, opened 2019-10-25
FIX=> Infinite Scroll issue fix on Orders > items search page

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2807

### What does this PR do?
BUG:  On Orders > Item search page on text change , search was not tirggering. 
Also, on removing all text at once was not triggering search text query. It has been fixed.

### DIFF
--- a/app/mixins/search_resource.js
+++ b/app/mixins/search_resource.js
@@ -44,7 +44,6 @@ export default Ember.Mixin.create({
   },
 
   onSearchTextChange: Ember.observer("searchText", function() {
-    const searchTextLength = this.get("searchText").length;
     if (isValidTextLength()) {
       this.reloadResults();
     }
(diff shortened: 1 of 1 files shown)


app/mixins/search_resource.js
```diff
@@ -44,12 +44,20 @@ export default Ember.Mixin.create({
   },
 
   onSearchTextChange: Ember.observer("searchText", function() {
-    if (this.get("searchText").length > this.get("minSearchTextLength")) {
-      this.rerenderOrShowNewResult();
+    const searchTextLength = this.get("searchText").length;
```
Darshan-Chauhan: What is the use of this const variable?

abhinav098: Yeah I think it's not required here @abulsayyad123 .

abulsayyad123: removed !!

PR #535, opened 2019-11-01
GCW-2804 support additional fields

abhinav098:
Ticket Link: https://jira.crossroads.org.hk/browse/GCW-2804
What does this PR do?
FEATURE: Adds the ability to add additional fields to the package

DESCRIPTION: StockIt has three extra tables related to items which contain fields that are only relevant to certain types of goods.
We want to duplicate this functionality for GoodCity

Impacted Areas
Package creation and package details

Mockup Link
https://xd.adobe.com/view/224f7314-f1c0-40b6-7d7a-929e616be6bb-547e/screen/7d129040-9142-4ed1-ba71-8011197fae58/Create-Item-Computer-form

Screenshots:
Create Page
Screen Shot 2019-10-22 at 11 24 31 AM

DropDown
Screen Shot 2019-10-22 at 11 27 34 AM

Detail page (Read only for now.)
Screen Shot 2019-10-22 at 11 27 57 AM


app/controllers/items/detail.js
```diff
@@ -2,210 +2,329 @@ import Ember from "ember";
...
+    }),
+
+    grades: Ember.computed("item.grade", function() {
       return [
-        ...records.rejectBy("state", "cancelled"),
-        ...records.filterBy("state", "cancelled")
+        { name: "A", id: "A" },
+        { name: "B", id: "B" },
+        { name: "C", id: "C" },
+        { name: "D", id: "D" }
```
swatijadhav: Remove this computed property as `GradeMixin` is already used.

app/components/additional-fields-input.js
```diff
@@ -0,0 +1,63 @@
...
+  store: Ember.inject.service(),
+  classNameBindings: ["inlineTextInput"],
+  inlineTextInput: false,
+  previousValue: "",
+
+  focusOut() {
+    const detailType = _.snakeCase(this.get("detailType")).toLowerCase();
+    const apiEndpoint = pluralize(detailType);
+    const detailId = this.get("detailId");
+    const url = `/${apiEndpoint}/${detailId}`;
```
patrixr: We want to avoid any **business logic/api** code within components as much as possible, because it makes them **unreusable** them in other pages/contexts.

They should be as **generic** as possible, an input field is just an input field (should not have hidden magic).
I recommend passing a callback from the controller to the component (e.g `onFocusOut=(action 'saveField' fieldName)`), so the logic lives outside of it.

So far other components have not followed those practices, which in turn forces us to copy them into many practically identical components. 

However:
There are some cases for some business logic components, usually big chunks of a page which are complex and not meant to be customised in any way
If you are in that situation, notice that I created a `components/goodcity/` folder within. 
That's what that is for, we prefix it by 'goodcity' to separate it from the others which are simply our own UI toolkit.
You can place a generic reusable component under `components/` and a custom one that extends it under `components/goodcity`


Apologies for the rant 😄

app/components/dropdown-search.js
```diff
@@ -0,0 +1,72 @@
...
+      this.send("setSelected", fieldName, newTag);
+    },
+
+    async setSelected(fieldName, value) {
+      this.get("onConfirm")(fieldName, value.tag);
+      if (this.get("displayPage")) {
+        const detailType = this.get("detailType").toLowerCase();
+        const apiEndpoint = pluralize(detailType);
+        const detailId = this.get("detailId");
+        const url = `/${apiEndpoint}/${detailId}`;
```
patrixr: Same as mentioned above. It applies generally to components

app/controllers/items/new.js
```diff
@@ -2,537 +2,609 @@ import Ember from "ember";
...
+    }),
+
+    grades: Ember.computed(function() {
+      return [
+        { name: "A", id: "A" },
+        { name: "B", id: "B" },
+        { name: "C", id: "C" },
+        { name: "D", id: "D" }
+      ];
+    }),
```
swatijadhav: Remove this, it's already part of `GradesMixin`

app/mixins/fetch_package_detail.js
```diff
@@ -0,0 +1,28 @@
+import Ember from "ember";
+import _ from "lodash";
+
+export default Ember.Mixin.create({
+  fetchPackageDetails: Ember.computed("packageDetails", function() {
```
patrixr: This is a mixin, but it doesn't look very re-usable.
It makes the assumption that the class I included it in has the following properties `packageDetails, showAdditionalFields insertFixedOption`. 

If it's just a mean of extracting code outside of a controller then I'd suggest helper classes.

Maybe something like `utils/subform_details.js`

```javascript
// utils/subform_details.js
export function buildSubformModel(c, ...) {
   ...
}

// controller
{
    fetchPackageDetails: Ember.computed("packageDetails", function() {
         helper.buildSubformModel( ... )
    }
}
```

If it actually is meant to be used in multiple places, then I suggest documenting how to integrate it with a comment. Basically setting default values to  the required properties and explaining in a comment that this mixin works by reading X,Y,Z props

Side note: The name `fetchPackageDetails` also sounds like a function, bit confusing, how about simply `subFormData` ?

app/mixins/grades_option.js
```diff
@@ -0,0 +1,12 @@
+import Ember from "ember";
```
patrixr: To complement what I wrote above, here is a much better use of a mixin. Many controllers might require grades, and I can include it anywhere and it'll work immediatly

app/mixins/search_resource.js
```diff
@@ -49,6 +50,28 @@ export default Ember.Mixin.create({
     }
   }),
 
+  onSearchCountry(searchText) {
```
patrixr: Here is a reusable mixin (well it tries to be, some of the comments I mentioned above apply here as well).

So we can't have any custom code of a controller. A mixin is not designed to stash away code, it should be a tool you integrate in the piece you're trying to build and used there to help you.

Notice how `onSearchCountry` conflicts in behaviours with `onSearchTextChange`, they have similar a purpose (triggering a search on change), but it doubles up. Which is a red flag.

@abulsayyad123 did a lot of integrations of search with this, he'll know best how to use it

app/routes/items/detail.js
```diff
@@ -25,6 +26,11 @@ export default AuthorizeRoute.extend({
       await Ember.RSVP.all(promises);
     }
 
+    let detail_type = model.get("detailType");
```
patrixr: Could you use camel case for the variables, we're all having a hard time transitioning between ruby and javascript 🤣 
I'm very guilty of that as well, (especially due to my hatred of camel case haha)

app/routes/items/detail.js
```diff
@@ -87,6 +101,25 @@ export default AuthorizeRoute.extend({
     return Ember.RSVP.all(ids.map(id => this.loadImage(id)));
   },
 
+  // loads package subform detail
+  async preLoadDetail(detail_type, detail_id) {
+    if (detail_type) {
+      return (
+        this.store.peekRecord(
```
patrixr: I think we have a `loadIfAbsent` helper somewhere, would help shorten this function down. I think prettier had its bit of fun 😄

app/routes/items/new.js
```diff
@@ -48,11 +48,21 @@ export default AuthorizeRoute.extend({
     }
   },
 
+  isAllowed(selectedSubform) {
```
patrixr:
```suggestion
isSubformAllowed(selectedSubform) {
```

app/routes/items/new.js
```diff
@@ -73,10 +83,31 @@ export default AuthorizeRoute.extend({
         controller.set("length", null);
         controller.set("width", null);
         controller.set("height", null);
-        controller.set("selectedGrade", { name: "B", id: "B" });
-        controller.set("selectedCondition", { name: "Used", id: "U" });
+        controller.set("selectedGrade", {
+          name: "B",
+          id: "B"
+        });
+        controller.set("selectedCondition", {
```
patrixr: Careful for the conditions, a while ago we replaced conditions by the actual lookup table
```javascript
conditions: Ember.computed(function() {
    return this.get("store").peekAll("donor_condition");
  })
```
So id `"U"` wouldn't match anything anymore

app/services/insert-fixed-option.js
```diff
@@ -0,0 +1,24 @@
+import Ember from "ember";
```
patrixr: Maybe worth moving that into the subform service ?  I have trouble understanding the name `insert-fixed-option` 🤔

app/services/subform-detail-service.js
```diff
@@ -0,0 +1,41 @@
+import Ember from "ember";
+import AjaxPromise from "./../utils/ajax-promise";
+import _ from "lodash";
+const { getOwner } = Ember;
+
+export default Ember.Service.extend({
```
patrixr: Good move creating a service for everything details related.

Note that there is a handy base API service you can use : 

```javascript
export default ApiBaseService.extend({
   ...

   updateRequest(paramsObj, previousValue) {
       this.PUT({ data })
    }
})
```

as well as an `async` mixin for everything related to the loading spinner : )

abhinav098:
Thanks for the valuable feedback @patrixr . I'll collaborate with @akshaygwd and help him do these changes. :+1:

app/routes/items/detail.js
akshaygwd: loadIfAbsent is present in orders_detail route its not present as a helper function,
I can move it out to somewhere else so that it would be accessible anywhere.

app/mixins/search_resource.js
akshaygwd: I am using a ember searchable select plugin for dropdown and it has its own **On search** method call. so can't use  **onSearchTextChange** observer.
I will create a separate mixin for this as we have to use search country at create page and update page as well.

app/routes/items/detail.js
patrixr: Sounds good :)

app/mixins/search_resource.js
patrixr: Your search method could trigger the onSearchTextChange. 
Although this mixing was built with infinite scroll in mind, if it's a drop-down we might not want to use it.

And again, a mixin is only appropriate if it can be re-used in multiple parts of the app. If not you can just place your methods in your controller

akshaygwd: I moved it in a mixin as it is used in create subform page and update subform page.

app/mixins/fetch_package_detail.js
akshaygwd: it is not possible to move it in a Utils folder as its not a ember class and services are not accessible over there

app/components/dropdown-search.js
```diff
@@ -0,0 +1,84 @@
...
+        const updateResponse = await this.get("onSetValue")(config);
+        let selectedValuesObj = {
+          ...this.get("selectedValuesDisplay")
+        };
+        const subformType = Object.keys(updateResponse)[0];
+        selectedValuesObj[fieldName] = {
+          id: this.isfixedDropdown(fieldName)
+            ? updateResponse[subformType][snakeCaseKey]
+            : updateResponse.id,
+          tag: this.isfixedDropdown(fieldName)
```
swatijadhav: `this.isfixedDropdown(fieldName)` is called multiple times within same function. Can we save that value in a variable and use it?

app/models/lookup.js
```diff
@@ -0,0 +1,9 @@
+import Model from "ember-data/model";
```
patrixr: @abhinav098 @akshaygwd  I'm probably a bit late on that end. But any possibilities of renaming the table to `form-lookup` (although that belongs more on the api PR than here).

If it's a lot of effort then carry on as it is, I just think it's more descriptive and understandable for anyone that might join the project later on 😄

app/mixins/fetch_package_detail.js
```diff
@@ -0,0 +1,38 @@
+import Ember from "ember";
```
patrixr: Similar to my previous comments. The usage of components and mixins are a bit misunderstood.
Given the time constraints we can move forward with this.
But keep it in mind next time you have an opportunity to refactor, feel free to ping me if you have questions :)

app/components/dropdown-search.js
```diff
@@ -0,0 +1,134 @@
...
+  fixedDropdownArr: ["frequency", "voltage", "compTestStatus", "testStatus"],
+  fixedDropdownArrId: [
+    "frequency_id",
+    "voltage_id",
+    "compTestStatus_id",
+    "testStatus_id"
+  ],
+
+  displayLabel: Ember.computed("addAble", function() {
+    return this.get("addAble") ? "Add New Item" : "";
```
patrixr: Might want to throw that string in the localisation file :)

app/controllers/items/detail.js
```diff
@@ -2,204 +2,345 @@ import Ember from "ember";
...
+        }
+        let selectedValues = this.get("item.detail.data");
+        let returnData = this.returnSelectedValues(selectedValues);
+        this.set("dataObjnew", {
+          ...returnData
+        });
+        return {
+          ...this.get("dataObjnew")
+        };
+      },
```
patrixr: I'm assuming you're using the spread operator to create a "clone" and to avoid changes to propagate to `dataObjnew` (very vague name btw, might be worth changing)

Since the `returnData`  ref is not used after being copied into `dataObjnew`, I don't know if the second use of the spread operator is needed. You might as well return it directly : )

app/controllers/items/new.js
```diff
@@ -2,537 +2,609 @@ import Ember from "ember";
...
           })
-          .then(data => {
-            if (this.get("isMultipleCountPrint")) {
-              this.printBarcode(data.item.id);
+          .then(() => this.hideLoadingSpinner());
+      },
+
+      deleteUnusedImage() {
+        if (this.get("imageKeys.length")) {
+          new AjaxPromise(
```
patrixr: might be worth moving to a service when there's time

app/controllers/items/new.js
```diff
@@ -2,537 +2,609 @@ import Ember from "ember";
...
+            { cloudinary_id: this.get("imageKeys") }
+          );
+          this.set("imageKeys", "");
+          this.set("newUploadedImage", null);
+        }
+      },
+
+      cancelForm() {
+        this.get("messageBox").custom(
+          "You will lose all your data. Are you sure you want to cancel this item?",
```
patrixr: Translation :)

app/routes/items/new.js
```diff
@@ -73,10 +90,40 @@ export default AuthorizeRoute.extend({
...
+          name: "Used",
+          id: "U"
```
patrixr: Still don't understand how an id of `"U"`works

abhinav098: It's an existing piece of code @patrixr, We need to check how this U is handled on the API side.

app/styles/app.scss
```diff
@@ -18,6 +18,7 @@
 @import "goods_details";
 @import "appointment_details";
 
+@import "node_modules/ember-searchable-select/app/styles/ember-searchable-select/style";
```
patrixr: I think sass node_modules should be imported from ember-cli-build.js ... but I'm not sure
@Darshan-Chauhan would you know ?

Darshan-Chauhan: @patrixr , I agree. I would prefer third party imports in `ember-cli-build`.

app/routes/items/new.js
abhinav098: I checked about that selectedCondition. It's not used anywhere in the `new.js` controller or `new.hbs` file. We can remove it.

app/components/additional-fields-input.js
```diff
@@ -0,0 +1,48 @@
+import Ember from "ember";
+import config from "stock/config/environment";
+import { pluralize } from "ember-inflector";
+import _ from "lodash";
+
+export default Ember.TextField.extend({
+  tagName: "input",
```
Darshan-Chauhan: This file looks like much similar to other components.
Can we use them or refactor?
I think the only difference is classNameBinding which we can pass it to the component as parameter.

app/components/dropdown-search.js
```diff
@@ -0,0 +1,134 @@
...
+export default Ember.Component.extend({
+  selected: [],
+  previousValue: "",
+  store: Ember.inject.service(),
+  subformDetailService: Ember.inject.service(),
+  resourceType: Ember.computed.alias("packageDetails"),
+  fixedDropdownArr: ["frequency", "voltage", "compTestStatus", "testStatus"],
+  selectedData: Ember.computed.alias("selectedValue"),
+  selectedDataDisplay: Ember.computed.alias("selectedValuesDisplay"),
+  fixedDropdownArr: ["frequency", "voltage", "compTestStatus", "testStatus"],
```
Darshan-Chauhan: This is repeated. Line 12: https://github.com/crossroads/stock.goodcity/pull/535/files#diff-0046bcc5c5d0932112242a81a9ea80c1R12

app/components/dropdown-search.js
```diff
@@ -0,0 +1,134 @@
...
+
+export default Ember.Component.extend({
+  selected: [],
+  previousValue: "",
+  store: Ember.inject.service(),
+  subformDetailService: Ember.inject.service(),
+  resourceType: Ember.computed.alias("packageDetails"),
+  fixedDropdownArr: ["frequency", "voltage", "compTestStatus", "testStatus"],
+  selectedData: Ember.computed.alias("selectedValue"),
+  selectedDataDisplay: Ember.computed.alias("selectedValuesDisplay"),
```
Darshan-Chauhan: Why do we need to use an alias here? I think we can directly use `selectedValuesDisplay`

app/components/dropdown-search.js
```diff
@@ -0,0 +1,134 @@
...
+const { getOwner } = Ember;
+
+export default Ember.Component.extend({
+  selected: [],
+  previousValue: "",
+  store: Ember.inject.service(),
+  subformDetailService: Ember.inject.service(),
+  resourceType: Ember.computed.alias("packageDetails"),
+  fixedDropdownArr: ["frequency", "voltage", "compTestStatus", "testStatus"],
+  selectedData: Ember.computed.alias("selectedValue"),
```
Darshan-Chauhan: Why do we need to use an alias here? I think we can directly use `selectedValue`

app/components/dropdown-search.js
```diff
@@ -0,0 +1,134 @@
...
+    "voltage_id",
+    "compTestStatus_id",
+    "testStatus_id"
+  ],
+
+  displayLabel: Ember.computed("addAble", function() {
+    return this.get("addAble") ? "Add New Item" : "";
+  }),
+
+  selectedOptionDisplay: Ember.computed("dropDownValues", function() {
```
Darshan-Chauhan: I am not able to understand the names here
`selectedOptionDisplay` and  `selectedDataDisplay`.. It's very confusing

app/components/dropdown-search.js
```diff
@@ -0,0 +1,134 @@
...
+  returnConfig(value, fieldName) {
+    const isfixedDropdown = this.isfixedDropdown(fieldName);
+    return {
+      value: isfixedDropdown ? value.id : value.tag,
+      name: isfixedDropdown ? `${fieldName}_id` : fieldName,
+      previousValue: this.get("previousValue")
+    };
+  },
+
+  returnsnakeCaseKey(fieldName) {
```
Darshan-Chauhan:
```suggestion
  returnSnakeCaseKey(fieldName) {
```

app/components/dropdown-search.js
```diff
@@ -0,0 +1,134 @@
...
+    const isfixedDropdown = this.isfixedDropdown(fieldName);
+    return isfixedDropdown
+      ? _.snakeCase(`${fieldName}_id`)
+      : _.snakeCase(fieldName);
+  },
+
+  retunLabelEn(responseKey) {
```
Darshan-Chauhan: Typo
```suggestion
  returnLabelEn(responseKey) {
```

app/components/dropdown-search.js
```diff
@@ -0,0 +1,134 @@
...
+      .get("labelEn");
+  },
+
+  returnTag(responseKey, fieldName) {
+    const isfixedDropdown = this.isfixedDropdown(fieldName);
+    return isfixedDropdown ? this.retunLabelEn(responseKey) : responseKey;
+  },
+
+  actions: {
+    addNew(fieldName, text) {
```
Darshan-Chauhan:
```suggestion
    addNewField(name, text) {
```

app/mixins/grades_option.js
```diff
@@ -0,0 +1,12 @@
...
+import _ from "lodash";
```
Darshan-Chauhan: I don't think this is used

app/styles/app.scss
akshaygwd: @patrixr @Darshan-Chauhan i moved above import to ember-cli-build and added extension .scss and restarted the server after that css is not getting applied to dropdown fields

akshaygwd: So keeping it in the same file only.

app/components/additional-fields-input.js
akshaygwd: @Darshan-Chauhan for sending data for update i am creating an object and other files has keyup, whichkey, keydown, function which we don't required over here

PR #545, opened 2019-11-13
Clicking on an orders_packages navigates to order detail

patrixr:
### Ticket Link: 

Quick addition to https://jira.crossroads.org.hk/browse/GCW-2779

### What does this PR do?

When clicking on an orders_package block on the Item details page, it redirects to the order details

### Impacted Areas

Item details > publishing tab > orders_package block

### Additions

Created a `back-link` component which uses the default browser back navigation. We've long known that hardcoding the back links it not acceptable, so hopefully this is a stepping stone

```
{{#back-link}}
   <div> Click me to go back ! </div>
{{/back-link}}
```

### DIFF
--- a/app/templates/components/goodcity/orders-package-block.hbs
+++ b/app/templates/components/goodcity/orders-package-block.hbs
@@ -3,7 +3,7 @@
     <div>{{fa-icon (orders-package-icon orderPkg.state) size='lg'}}</div>
     <div>{{orderPkg.quantity}}</div>
   </div>
-  <div class="columns small-9 sub-block main {{ orderPkg.designation.state }}">
+  <div class="columns small-9 sub-block main {{ orderPkg.designation.state }}" {{action 'redirectToOrderDetail' orderPkg.designation.id}}>
     <div class="row">
       <div class="columns small-11 text-align-left">
         {{fa-icon orderPkg.designation.transportIcon size='lg'}}
--- a/app/templates/orders/_orders_back_link_path.hbs
+++ b/app/templates/orders/_orders_back_link_path.hbs
@@ -19,16 +19,11 @@
       <i class="back_icon" aria-hidden="true">{{fa-icon 'angle-left'}}</i>
       <div>{{t "back"}}</div>
     {{/link-to}}
-  {{else if (is-or filterService.getOrderStateFilters.length filterService.getOrderTypeFilters.length)}}
-      {{#link-to "orders.index" (query-params isFiltered=true) classNames="back" }}
-        <i class="back_icon" aria-hidden="true">{{fa-icon 'angle-left'}}</i>
-        <div>{{t "back"}}</div>
-      {{/link-to}}
   {{else}}
-    {{#link-to "orders.index" classNames="back" }}
+    {{#back-link}}
       <i class="back_icon" aria-hidden="true">{{fa-icon 'angle-left'}}</i>
       <div>{{t "back"}}</div>
-    {{/link-to}}
+    {{/back-link}}
   {{/if}}
 </section>
 
--- a/app/components/goodcity/orders-package-block.js
+++ b/app/components/goodcity/orders-package-block.js
@@ -198,5 +198,11 @@ export default Ember.Component.extend(AsyncMixin, {
         })
       );
     }
-  )
+  ),
+
+  actions: {
+    redirectToOrderDetail: function(orderId) {
+      this.router.transitionTo("orders.detail", orderId);
+    }
+  }
 });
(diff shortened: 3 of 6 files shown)


Darshan-Chauhan:
@patrixr , this is nice.
Would it be better to add optional params `path` and `id` so we can navigate to any page?

app/templates/orders/_orders_back_link_path.hbs
```diff
@@ -19,16 +19,11 @@
...
-        <i class="back_icon" aria-hidden="true">{{fa-icon 'angle-left'}}</i>
-        <div>{{t "back"}}</div>
-      {{/link-to}}
   {{else}}
-    {{#link-to "orders.index" classNames="back" }}
+    {{#back-link}}
       <i class="back_icon" aria-hidden="true">{{fa-icon 'angle-left'}}</i>
       <div>{{t "back"}}</div>
-    {{/link-to}}
+    {{/back-link}}
```
patrixr: I couldn't bring myself to add one more conditional here. It's too bad a practice, so attempting something a bit different.

Changes:
- Removed the first `orders.index` link because `isFiltered=true` doesn't seem to be used anywhere. Probably some old code
- Used a back-link as a "default" behaviour

Questions: 

What are the cases in which we **don't** want to navigate back to where we came from. I think a history.back would actually handle a lot of the cases written above.
I think that starting by finding the exceptions, and leaving the rest to default is a good start. Rather than listing the actual expected navigations one by one.

app/templates/components/goodcity/orders-package-block.hbs
```diff
@@ -3,7 +3,7 @@
     <div>{{fa-icon (orders-package-icon orderPkg.state) size='lg'}}</div>
     <div>{{orderPkg.quantity}}</div>
   </div>
-  <div class="columns small-9 sub-block main {{ orderPkg.designation.state }}">
+  <div class="columns small-9 sub-block main {{ orderPkg.designation.state }}" {{action 'redirectToOrderDetail' orderPkg.designation.id}}>
```
patrixr: Note: this is the actual change of this PR

patrixr:
> @patrixr , this is nice.
> Would it be better to add optional params `path` and `id` so we can navigate to any page?

@Darshan-Chauhan where are you suggesting we add them ? could you comment on the code so that I can see 😄

Darshan-Chauhan:
@patrixr , something like this.

```
{{#back-link path="items" id="123"}}
   <div> Click me to go back ! </div>
{{/back-link}}
```

Here `path` and `id` both are optional as path can be `items.index` where `id` is not needed.

Just a thought. What do you think?

patrixr:
> @patrixr , something like this.
> 
> ```
> {{#back-link path="items" id="123"}}
>    <div> Click me to go back ! </div>
> {{/back-link}}
> ```
> 
> Here `path` and `id` both are optional as path can be `items.index` where `id` is not needed.
> 
> Just a thought. What do you think?

I see where you're going with this. But that case would actually result in a forward navigation, meaning it wouldn't pop the history stack. In which case it results back to using `link-to`

I'm trying to really differentiate those 2 things, for more "awareness".

We could modify `back-link` as you're suggesting, but doing a `replaceRoute`, but that will likely lead to having the same page twice in a row in the history stack if whoever uses is not aware of how it works. Here as well using `link-to` with `replace=true` fits.

Just thinking out loud. But I think a native back action works most of the time, I personally always use the browser's back button

Darshan-Chauhan:
@patrixr , Hmmm I agree. `replace=true` would be a better choice.
I was thinking about solving back link issue where we've lots of `{{link-to}}` with if-else.

Let's keep `{{back-link}}` as it is, we can figure out some generic solution for `{{link-to}}` with if-else.

patrixr:
> @patrixr , Hmmm I agree. `replace=true` would be a better choice.
> I was thinking about solving back link issue where we've lots of `{{link-to}}` with if-else.
> 
> Let's keep `{{back-link}}` as it is, we can figure out some generic solution for `{{link-to}}` with if-else.

Cool. I actually think the `if/else` could be replaced with a native back (to be tested).
That is, if we manage the exceptions. 
The only one I can think of for orders.details, is when we navigate from the order creation process, in which case we don't want to go back to it.
IF that order creation process did a `replace` when navigating to the order.details, when the back navigation would also work.

But it's hard to know the actual use cases 😢

Darshan-Chauhan:
> But it's hard to know the actual use cases

We'll have to talk with Matt and find out.

patrixr:
>
> We'll have to talk with Matt and find out.

Yup, I'll merge this as it is. There will likely be some slight changes in behaviours, but unlikely to cause much damage

PR #550, opened 2019-11-21
Refactor move

patrixr:
This PR goes hand in hand with this [backend change](https://github.com/crossroads/api.goodcity/pull/854) (read first for context)

We want to start using the new Order fulfilment "operations" (as they've been called on the api).
That will allow a smooth(er) transition to a stockit free world later on since those have been implemented with that in mind.

As a result, here are the changes that occurred : 

- Implementing the proper "move" workflow, which assumes multi-location and partial quantity is possible. (modifying the quantity is locked for the time being, but fully functional)
- removing any code that refers to old endpoints or previous partial quantity implementation


app/mixins/move_actions.js
```diff
@@ -0,0 +1,137 @@
...
+
+    return this.get("locationService").userPickLocation({
+      headerText: text,
+      presetLocations: presetLocations
+    });
+  },
+
+  quantityAtSource() {
+    const pkg = this.get("moveTarget");
+    const source = this.get("moveFrom");
```
namrataukirde: How about naming it as `sourceLocation`?

app/services/location-service.js
```diff
@@ -21,6 +23,20 @@ export default ApiBaseService.extend(NavigationAwareness, {
     this.getWithDefault("onLocationSelected", _.noop)(null);
   },
 
+  async movePackage(pkg, opts = {}) {
+    const { from, to, quantity } = opts;
+
+    const payload = await this.PUT(`/packages/${pkg.get("id")}/move`, {
```
namrataukirde: Is there any scenario where we need to show an error to users if any move operation fails?

app/mixins/move_actions.js
patrixr: mostly for consistency with the backend (more aesthetically pleasing too imo)

app/services/location-service.js
patrixr: There are, but not in the service. An exception coming from the service is a perfectly valid outcome.

Whoever calls the service has to handle the error, that would usually be the page.
Right now it's letting the default error handler show the popup, but I do have an improvement in mind regarding that : )

PR #551, opened 2019-11-22
FIX=> On refresh something went wrong error popsup during PackageType Selection.

abulsayyad123:
### What does this PR do?

BUG: On refresh something went wrong error popsup during PackageType Selection. 

### Impacted Areas
Package Type selection during Package creation process.

Thanks.

### DIFF
--- a/app/components/focus-textfield.js
+++ b/app/components/focus-textfield.js
@@ -37,9 +37,10 @@ export default Ember.TextField.extend({
   },
 
   didInsertElement() {
-    var routes = this.router.router.currentHandlerInfos;
-    var currentRoute = routes.pop();
-    var isIndexRoute = currentRoute.name === "items.index" ? true : false;
+    let routes = this.router.router.currentHandlerInfos;
+    let currentRoute = routes && routes.pop();
+    let isIndexRoute =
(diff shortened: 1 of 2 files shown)


app/components/focus-textfield.js
```diff
@@ -37,9 +37,10 @@ export default Ember.TextField.extend({
   },
 
   didInsertElement() {
-    var routes = this.router.router.currentHandlerInfos;
-    var currentRoute = routes.pop();
-    var isIndexRoute = currentRoute.name === "items.index" ? true : false;
+    let routes = this.router.router.currentHandlerInfos;
+    let currentRoute = routes && routes.pop();
+    let isIndexRoute =
+      currentRoute && currentRoute.name === "items.index" ? true : false;
```
patrixr: I know this isn't the change you made. But the fact that a `textfield` modifies the routing history is deeply disturbing.

Would you know why it's doing that ? Or why it even needs routing info ?

abulsayyad123: You are right.. dealing with routing history in component isn't correct.I didn't get much chance to look the entire flow,  I just hot-fix for release sake.

PR #554, opened 2019-11-22
allow user creation without mobile

namrataukirde:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2838

BUG: 
The mobile field is an optional field on user creation form whereas not adding a mobile number throws invalid mobile number error.

Expected output: User should get created without mobile number.

ISSUE: 
const mobilePhone = mobilePhoneValue.length &&
      `${config.APP.HK_COUNTRY_CODE}${mobilePhoneValue}`;

In the above line, if mobile is blank then it returns 0 and that is send as mobile input to API and API throws an error.

Please Review.

Thanks.

### DIFF
--- a/app/controllers/users/new.js
+++ b/app/controllers/users/new.js
@@ -25,12 +25,16 @@ export default Ember.Controller.extend({
     this.set("preferredPhone", "");
   },
 
+  formatMobileNumber() {
+    const mobile = this.get("mobilePhone");
+    return mobile.length
+      ? config.APP.HK_COUNTRY_CODE + this.get("mobilePhone")
+      : mobile;
+  },
+
   getRequestParams() {
     const preferredPhoneValue = this.get("preferredPhone");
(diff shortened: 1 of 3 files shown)


steveyken:
Thanks! Would there be a similar bug on Browse?

app/controllers/users/new.js
```diff
@@ -25,12 +25,16 @@ export default Ember.Controller.extend({
     this.set("preferredPhone", "");
   },
 
+  formatMobileNumber() {
+    const mobile = this.get("mobilePhone");
+    return mobile.length
+      ? config.APP.HK_COUNTRY_CODE + this.get("mobilePhone")
+      : mobile;
+  },
+
```
patrixr: I'm not really seeing the difference with the previous version 🤔

namrataukirde: Previous code was returning 0, in this case, it will return blank.

PR #559, opened 2019-11-29
GCW-2900 clear on back

akshaygwd:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2900

### What does this PR do?

BUG: This PR clear all the subform attributes on clicking the cancel button.
Please review.
Thanks


app/controllers/items/new.js
```diff
@@ -131,6 +132,7 @@ export default GoodcityController.extend(
     },
 
     showPiecesInput: Ember.computed("codeId", function() {
+      this.clearSubformAttributes();
```
patrixr: An update inside a computed property is a big code smell.
Reading a boolean (`showPiecesInput `) should absolutely not modify any data.

You can call that function in your actions at the appropriate time instead :)

patrixr: or if it needs to react to a change, then better use an observer than a computed property

PR #561, opened 2019-12-02
GCW-2840-Edit subform feature

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2840

### What does this PR do?

FEATURE: Additional fields editing package type for packages already created. There are different scenarios for subform is handled in this PR.

BUG: On refresh on Item creation package do not load `subform` if packageType has `subform`

### Impacted Areas

Package details page where user can edit existing subform to same/other/no subforms.


app/services/package-service.js
```diff
@@ -15,5 +17,14 @@ export default ApiBaseService.extend({
 
   createPackage(pkgParams) {
     return this.POST(`/packages`, pkgParams);
+  },
+
+  updatePackage(pkgId, pkgParams) {
+    return this.PUT(`/packages/${pkgId}`, pkgParams);
+  },
+
+  deletePackage(detailType, detailId) {
```
abhinav098: @abulsayyad123 name is ambiguous, it's deleting package's detail and the method says as if it is deleting the package.

abhinav098: And i think, this method is related to subform detail, should go to `subform-detail-service.js`. Let me know your thoughts.

app/routes/items/new.js
```diff
@@ -69,78 +54,98 @@ export default AuthorizeRoute.extend({
...
 
-  async setupController(controller, model) {
+  setupController(controller, model) {
     this._super(controller, model);
+    const store = this.get("store");
+    this.initializeController();
+
+    if (this.get("newItemRequest")) {
+      this.initializeDimensions();
+      this.manageSubformDetails();
```
abhinav098: 👍

app/controllers/items/search_code.js
```diff
@@ -1,44 +1,112 @@
...
     } else {
-      return this.store.query('code', { stock: true });
+      return this.store.query("code", { stock: true });
     }
   }),
 
+  deleteAndAssignNew(packageType) {
+    const item = this.get("item");
+    const detailType = item.get("detailType");
+    const detailId = item.get("detail.id");
```
abhinav098: @abulsayyad123, you can do, `item.detailId` no need of `detail.id`

app/controllers/items/search_code.js
```diff
@@ -1,44 +1,112 @@
...
+      null,
+      translation.t("continue"),
+      () => {
+        this.deleteAndAssignNew(pkgType);
+      }
+    );
+  },
+
+  isSubformPackage(packageType) {
+    return !!packageType.get("subform");
```
abhinav098: @abulsayyad123 , it's better to check `["computer", "computer_accessory", "electrical"].indexOf(packageType.get("subform"))` because some of the packageTypes have subform field as "medical".

app/services/package-service.js
abulsayyad123: You are correct @abhinav098 about the naming. But my naming convention is according to first parameter. You can read `deletePackage(detailType)` as `deletePackageDetailType`.

app/controllers/items/search_code.js
```diff
@@ -1,44 +1,112 @@
...
+    this.get("packageService")
+      .deletePackage(detailType, detailId)
+      .then(() => {
+        if (this.isSubformPackage(packageType)) {
+          this.assignNew(packageType);
+        } else {
+          this.assignNew(packageType, { deleteDetailId: true });
+        }
+      });
+  },
```
patrixr: You can save a lot of lines by using async functions. This applies generally, just using this function as an example.
(for the spinner, there is an async mixin you can use )

```suggestion
async deleteAndAssignNew(packageType) {
    const detailType = item.get("item.detailType");
    const detailId = item.get("item.detailId");

    await this.runTask(this.get("packageService").deletePackage(detailType, detailId));
    this.assignNew(packageType, { deleteDetailId: this.isSubformPackage(packageType) });
},
```

app/controllers/items/search_code.js
```diff
@@ -1,44 +1,112 @@
...
+      package_type_id: type.get("id")
+    };
+    if (!this.isSamePackage(type)) {
+      packageParams.detail_type = capitalize(type.get("subform"));
+    }
+    if (deleteDetailId) {
+      packageParams.detail_id = null;
+    }
+    this.showLoadingSpinner();
+    this.get("packageService")
```
patrixr: Make a habit of always returning promises, because other people will likely assume it does. The entire javascript ecosystem is based on asynchronous code, so it's important to have that capability of "await-ing" the promise

patrixr: Also, since there is already a `packageService`, might as well move a lot of the Ajax mumbo-jumbo there : )

app/controllers/items/search_code.js
```diff
@@ -1,44 +1,112 @@
-import Ember from 'ember';
-import SearchCode from '../search_code';
-import AjaxPromise from 'stock/utils/ajax-promise';
-const { getOwner } = Ember;
+import Ember from "ember";
```
patrixr: Can we rename that page to something more meaningful ? e.g `assign_subform`

`search_code` is already (_wrongly_) used in many places, but I don't think we should accept that as a standard. As it suggests that this is a re-usable page used to search codes, which is not the case. It's a hardcoded page which does a very specific task. Which is why we have so many copy pasted pages.

I would normally suggest to make the page generic, but I know how deep we are in that bad practice, and refactoring it is a rough task. 

(FYI: I'm in the process of removing those search pages and replacing them with asynchronous animated overlays)

app/routes/items/new.js
```diff
@@ -69,78 +54,98 @@ export default AuthorizeRoute.extend({
...
+      this.setUpPackageImage();
+      window.localStorage.setItem("isSelectLocationPreviousRoute", false);
```
patrixr: I know this is probably old legacy code, but that is sadly not an acceptable use of `localStorage`, nor is it the right way to structure navigation. I suggest removing it and using a default browser back instead.

It might lead to inconsistencies/bugs but we'll fix those when they actually do happen. If we want to go from hell to heaven, we still need to walk through some fire 😄

app/routes/items/new.js
```diff
@@ -69,78 +54,98 @@ export default AuthorizeRoute.extend({
...
+  },
 
-    controller.set("inventoryNumber", this.get("inventoryNumber"));
-    controller.set("displayInventoryOptions", false);
-    controller.set("autoGenerateInventory", true);
-    controller.set("inputInventory", false);
+  initializeController() {
+    const controller = this.controller;
+    if (!controller.get("inventoryNumber")) {
+      controller.send("autoGenerateInventoryNumber");
```
patrixr: Since the inventory number generation is done on the backend, I'm pretty sure this could lead to asynchronous drama.

You can await the inventory number creation, set it, and then proceed

app/routes/items/new.js
```diff
@@ -69,78 +54,98 @@ export default AuthorizeRoute.extend({
...
-      if (codeId) {
-        let selected = this.get("store").peekRecord("code", codeId);
-        if (selected) {
-          let selectedSubform = selected.get("subform");
-          if (this.isSubformAllowed(selectedSubform)) {
-            controller.set("showAdditionalFields", true);
-            let details = await this.store.query(selectedSubform, {
-              distinct: "brand"
+  async manageSubformDetails() {
+    const controller = this.controller;
```
patrixr: Tip: to avoid endless nesting of IFs, you can use negatives and returns, which usually leads to simpler and more readable code

From
```javascript
if (A) {
  if (B) {
    if (C) {
        // logic
    }
  }
}
```

to 
```javascript
if (!A) return or throw;
if (!B) return or throw;
if (!C) return or throw;

// logic
```

app/routes/items/new.js
```diff
@@ -69,78 +54,98 @@ export default AuthorizeRoute.extend({
...
+    if (codeId) {
+      let selected = store.peekRecord("code", codeId);
+      if (selected) {
+        let selectedSubform = selected.get("subform");
+        if (this.isSubformAllowed(selectedSubform)) {
+          controller.set("showAdditionalFields", true);
+          let details = await store.query(selectedSubform, {
+            distinct: "brand"
+          });
+          if (this.get("transitionFrom") === "code") {
```
patrixr: We want to avoid those kinds of history lookups. It's not tolerant to change and prevents other usages of the same page.
If there's a specific behaviour the page needs to do, then the caller of the page can pass an argument to it.

Also, why should the page reset values only when it comes from 'code' ? It feels weird

app/routes/items/new.js
```diff
@@ -69,78 +54,98 @@ export default AuthorizeRoute.extend({
...
-      }
-      window.localStorage.setItem("isSelectLocationPreviousRoute", false);
+  setUpPackageImage() {
+    const controller = this.controller;
+    const imageKey = controller.get("imageKeys");
+    const store = this.store;
+    if (
+      imageKey &&
+      imageKey.length &&
+      window.localStorage.isSelectLocationPreviousRoute === "true"
```
patrixr: Same comment as above. Why an image only when we come from select location ? I can't justify that in anyways.

Logic is too convoluted, why can't a `if (packageHasImage) -> displayimage` logic work here

app/services/package-service.js
patrixr: @abulsayyad123 Imagine someone writes `deletePackage(123)` what would you assume that method does 😄 

I think I agree that the name is weird.

And even if written `deletePackage(detailType)`, it could be understood as `delete the package of the following detail`

However you do make a point that meaning can sometimes come from argument names, but in those scenarios it's better to use named argument. Ruby is really good for that, in JS that would be an object.

example: 

```javascript

// V1

function sendMail(body, author, recipient) { ... }
sendMail('hello', user1, user2); // We don't know who is the author unless we look at the source code of the function

function sendMail(body, context= {}) { ... }

sendMail('hello', { author: user1, recipient: user2 }); // I don't need to look at the source code to understand 
```

PR #562, opened 2019-12-04
Gcw 2853 support printer selection in ui

akshaygwd:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2853

### What does this PR do?

FEATURE: Add a "Printer" single select control at the bottom right of the create form (in both admin and stock apps)

Screenshots:
Screen Shot 2019-12-04 at 7 35 55 PM

Please review 
Thanks.

### DIFF
--- a/app/controllers/items/new.js
+++ b/app/controllers/items/new.js
@@ -52,7 +52,9 @@ export default GoodcityController.extend(
     isAllowedToPublish: false,
     imageKeys: Ember.computed.localStorage(),
     i18n: Ember.inject.service(),
+    session: Ember.inject.service(),
     packageService: Ember.inject.service(),
+    printerService: Ember.inject.service(),
     cancelWarning: t("items.new.cancel_warning"),
     selectedPrinter: [],
     displayFields: Ember.computed("code", function() {
@@ -69,22 +71,15 @@ export default GoodcityController.extend(
       return this.get("i18n").t(str);
     },
 
-    allAvailablePrinter: Ember.computed("availablePrinters", function() {
-      let printers = [];
-      this.get("availablePrinters").map(printer => {
-        let tag = printer.get("name");
-        printers.push({ id: printer.get("id"), tag: tag });
-      });
-      let userPrinter = this.store
-        .peekRecord("user", this.session.get("currentUser.id"))
-        .get("printer");
+    allAvailablePrinter: Ember.computed(function() {
+      let userPrinter = this.get("session").userDefaultPrinter();
       if (userPrinter) {
         this.set("selectedPrinter", {
           id: userPrinter.id,
           tag: userPrinter.get("name")
         });
       }
-      return printers;
+      return this.get("printerService").allAvailablePrinter();
     }),
 
     setLocation: Ember.observer("scanLocationName", function() {
--- a/app/controllers/items/new.js
+++ b/app/controllers/items/new.js
@@ -620,10 +620,10 @@ export default GoodcityController.extend(
       setPrinterValue(value) {
         console.log(this);
         this.set("selectedPrinterDisplay", {
-          id: value.id,
-          tag: this.get("store")
+          name: this.get("store")
             .peekRecord("printer", value.id)
-            .get("name")
+            .get("name"),
+          id: value.id
         });
       },
 
--- a/app/controllers/items/new.js
+++ b/app/controllers/items/new.js
@@ -612,7 +612,7 @@ export default GoodcityController.extend(
       },
 
       setPrinterValue(value) {
-        let printerId = value.id;
+        const printerId = value.id;
         this.set("selectedPrinterId", printerId);
         this.get("printerService").updateUserDefaultPrinter(printerId);
       },
--- a/app/controllers/items/new.js
+++ b/app/controllers/items/new.js
@@ -70,15 +70,15 @@ export default GoodcityController.extend(
       return this.get("i18n").t(str);
     },
 
-    allAvailablePrinter: Ember.computed(function() {
-      return this.get("printerService").allAvailablePrinter();
+    allAvailablePrinters: Ember.computed(function() {
+      return this.get("printerService").allAvailablePrinters();
     }),
 
     selectedPrinterDisplay: Ember.computed({
       get(key) {
         let userPrinter = this.get("session").userDefaultPrinter();
         if (!userPrinter) {
-          return [];
+          return;
         }
         return { id: userPrinter.id, tag: userPrinter.get("name") };
       },
--- a/app/controllers/items/new.js
+++ b/app/controllers/items/new.js
@@ -426,15 +426,6 @@ export default GoodcityController.extend(
         });
     },
 
-    updateUserDefaultPrinter(printerId) {
-      new AjaxPromise(
-        `/users/${this.get("session.currentUser.id")}`,
-        "PUT",
-        this.get("session.authToken"),
-        { user: { printer_id: printerId } }
-      ).then(data => this.get("store").pushPayload(data));
-    },
-
     actions: {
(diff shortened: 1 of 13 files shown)


app/controllers/items/new.js
```diff
@@ -69,6 +70,23 @@ export default GoodcityController.extend(
       return this.get("i18n").t(str);
     },
 
+    allAvailablePrinter: Ember.computed(function() {
+      return this.get("printerService").allAvailablePrinter();
```
abhinav098: @akshaygwd `allAvailablePrinter` should be `allAvailablePrinters` because its gonna return an array of printers.

abhinav098: Same comment for this computed property.

app/controllers/items/new.js
```diff
@@ -69,6 +70,23 @@ export default GoodcityController.extend(
...
+    }),
+
+    selectedPrinterDisplay: Ember.computed({
+      get(key) {
+        let userPrinter = this.get("session").userDefaultPrinter();
+        if (!userPrinter) {
+          return [];
```
abhinav098: @akshaygwd , I think, in case of no default printer it should send the `firstObject` from the printers for that location as discussed in the call once. Because sending a blank array will display nothing on the printer selection dropdown.

app/controllers/items/new.js
```diff
@@ -360,8 +378,12 @@ export default GoodcityController.extend(
 
     printBarcode(packageId) {
       const labels = this.get("labels");
-      this.get("packageService")
-        .printBarcode({ package_id: packageId, labels })
+      return this.get("packageService")
```
abhinav098: nice use of `packageService` 👍

app/controllers/items/new.js
```diff
@@ -593,6 +615,15 @@ export default GoodcityController.extend(
         });
       },
 
+      setPrinterValue(value) {
+        this.set("selectedPrinterDisplay", {
+          id: value.id,
+          tag: this.get("store")
```
abhinav098:
```suggestion
          tag: this.get("selectedPrinterDisplay").tag
```
No need of peekRecord. You should use the property that you defined above viz `selectedPrinterDisplay`.

app/services/printer-service.js
```diff
@@ -0,0 +1,14 @@
+import Ember from "ember";
+
+export default Ember.Service.extend({
+  store: Ember.inject.service(),
+  allAvailablePrinter() {
```
abhinav098:
```suggestion
  allAvailablePrinters() {
```

app/services/session.js
```diff
@@ -17,6 +17,11 @@ export default Ember.Service.extend({
     return store.peekAll("user_profile").get("firstObject") || null;
   }).volatile(),
 
+  userDefaultPrinter: function() {
+    const store = this.get("store");
+    return store.peekRecord("user", this.get("currentUser.id")).get("printer");
```
abhinav098:
```suggestion
    return this.get("store").peekRecord("user", this.get("currentUser.id")).get("printer");
```

No need of storing store to a variable.

app/services/session.js
```diff
@@ -17,6 +17,11 @@ export default Ember.Service.extend({
     return store.peekAll("user_profile").get("firstObject") || null;
   }).volatile(),
 
+  userDefaultPrinter: function() {
+    const store = this.get("store");
```
abhinav098:
```suggestion
```

app/controllers/items/new.js
```diff
@@ -410,6 +432,19 @@ export default GoodcityController.extend(
         });
     },
 
+    updateUserDefaultPrinter(printerId) {
+      let userRecord = this.store.peekRecord(
+        "user",
+        this.get("session.currentUser.id")
+      );
```
swatijadhav: Unused variable declaration.

app/services/printer-service.js
```diff
@@ -0,0 +1,14 @@
...
+
+export default Ember.Service.extend({
+  store: Ember.inject.service(),
+  allAvailablePrinters() {
+    let availablePrinters = this.get("store").peekAll("printer");
+    let printers = [];
+    availablePrinters.map(printer => {
+      let tag = printer.get("name");
+      printers.push({ name: tag, id: printer.get("id") });
+    });
```
swatijadhav:
```js
    availablePrinters.map(printer => {
      printers.push({ 
         name: printer.get("name"), 
         id:   printer.get("id") 
      });
    });
```

app/styles/templates/items/_new.scss
```diff
@@ -149,7 +149,7 @@
   input,
   select {
     padding: 0 0.5rem;
-    background: #002352 !important;
+    background-color: #002352 !important;
```
swatijadhav: Use color variable name.

app/styles/templates/items/_new.scss
```diff
@@ -267,6 +267,44 @@
...
+
+        .dropdown-wrap {
+          .select {
+            select {
+              background-image: url(../assets/images/dropdown_icon.png);
+              background-repeat: no-repeat;
+              background-size: 1.2rem;
+              background-position: 95%;
+              font-size: 0.8rem;
+              background-color: #002352;
```
swatijadhav: Use color variable.

app/controllers/items/new.js
```diff
@@ -69,6 +70,23 @@ export default GoodcityController.extend(
...
+
+    selectedPrinterDisplay: Ember.computed("inventoryNumber", {
+      get(key) {
+        let userPrinter = this.get("session").userDefaultPrinter();
+        if (!userPrinter) {
+          return;
+        }
+        return { name: userPrinter.get("name"), id: userPrinter.id };
+      },
+      set(key, value) {
```
patrixr: The usage of a setter (which doesn't seem to set anything anyways) in combination with a computed property which reacts to a property `inventoryNumber` is definitely a code smell.

The only way it would make sense is if the `set` method actually modifies the `inventoryNumber` which in turn triggers the property to update.

I would suggest removing the setter, and having the property react to `selectedPrinterId` which you can set whenever you need to

`this.set('selectedPrinterDisplay', { ... })` This is dangerous, it defeats the whole purpose of the property computation 🔴 

`this.set('selectedPrinterId', 123)` this doesn't enforce anyone setting it to know that the format of the object is {name,id}, once set the computed property can react to it 💚

app/controllers/items/new.js
```diff
@@ -595,6 +630,17 @@ export default GoodcityController.extend(
...
       },
 
+      setPrinterValue(value) {
+        let printerId = value.id;
+        this.set("selectedPrinterDisplay", {
+          name: this.get("store")
+            .peekRecord("printer", printerId)
+            .get("name"),
+          id: printerId
+        });
```
patrixr: This `set` is the one I'm referring to in the comment above. It doesn't make sense since you're already building that object ( {name, id} ) in the computation

```suggestion
this.set('selectedPrinterId', printerId);
```

app/controllers/items/new.js
```diff
@@ -410,6 +432,19 @@ export default GoodcityController.extend(
         });
     },
 
+    updateUserDefaultPrinter(printerId) {
+      let userRecord = this.store.peekRecord(
+        "user",
+        this.get("session.currentUser.id")
+      );
+      new AjaxPromise(
```
patrixr: Feel free to move this into a service : )

app/services/printer-service.js
patrixr: if it's using `map`, then there's no need to push anything into an array, you can just return it : )

```javascript
  return availablePrinters.map(printer => printer.getProperties('name', 'id'))
```

PR #564, opened 2019-12-06
String sanitization prevents chinese characters in the organisation search

patrixr:
Remove reference to old (and soon to disappear) `search_module`

This search had been refactored with the new infinite list, but the reference to the old search_module was still there. And it was trimming the chinese characters out of the search text.

I simply removed that reference which is not needed.


steveyken:
:+1:

PR #565, opened 2019-12-10
Refactor designation

patrixr:
These frontend changes come hand in hand with : https://github.com/crossroads/api.goodcity/pull/874

## Changes

* Choosing an order to designate to is done via an overlay rather than a transition
* Deletion of pages no longer used
* Designation is done through new endpoint, allowing to request a specific quantity (currently locked to full quantity)
* On the orders_package blocks, the `edit quantity` option is now functional
* Addition of `goodcity_settings` (copy-pasta from Browse) with the following keys 
    - `stock.allow_partial_operations` (defaults to false)
    - `stock.only_publish_singletons` (defaults to true)


app/locales/zh-tw/translations.js
```diff
@@ -760,7 +758,7 @@ export default {
       "增加{{state}}的數量到{{totalQty}} 對於{{designationCode}}! ({{qty}} 原本{{pkgState}}-物件{{inventory number}})",
     warning_text_without_link:
       "增加{{state}}的數量到{{totalQty}} 對於{{orderCode}}! ({{qty}} 原本{{designatedState}})",
-    designate_to: "指定這組物件到{{orderCode}}",
+    designate_to: "Designate this package to:",
     designate_entire: " 指定全數",
     designate_part: "指定部份"
```
abhinav098: `quantity_input: "Quantity to designate"` missing in chinese file.

app/mixins/move_actions.js
```diff
@@ -23,31 +21,27 @@ const ALLOW_PARTIAL_QTY = config.APP.ALLOW_PARTIAL_OPERATIONS;
...
-        }
-
-        this.set("_moveQty", qty);
-        return qty;
+  moveQty: Ember.computed("_moveQty", {
+    get(k) {
+      return this.get("_moveQty");
+    },
+    set(k, value) {
+      const allowPartial = this.get("settings.allowPartialOperations");
```
abhinav098: can we use `editableQty` here instead ? as it is already set above.

patrixr: We could but I'd rather not, I left it that way on purpose for readability.

Reading `if notEditable then checkIsPartial` doesn't make sense
whereas reading `if !allowPartial then checkIsPartial`  does.

Even if both variables come from the same source, they hold different meanings in their names :)

app/locales/zh-tw/translations.js
patrixr: good eye 👍

PR #566, opened 2019-12-11
Gcw 2774 support creating boxes or pallets

akshaygwd:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2774

### What does this PR do?

FEATURE: 
This ticket is for creating an empty box or pallet. It does not include adding or removing things from box/pallets. Those features will be implemented as soon as our new architecture for inventory_packages is in place.

##Screenshots
Screen Shot 2019-12-11 at 5 03 32 PM

Screen Shot 2019-12-11 at 5 03 44 PM

Screen Shot 2019-12-26 at 7 11 55 PM


app/controllers/app_menu_list.js
```diff
@@ -1,44 +1,55 @@
-import Ember from 'ember';
-import config from '../config/environment';
+import Ember from "ember";
+import config from "../config/environment";
```
swatijadhav: Use an absolute path.

app/components/goodcity/code-search-overlay.js
```diff
@@ -0,0 +1,133 @@
...
+    this._super("code-search-overlay");
+  },
+
+  allPackageTypes: Ember.computed("open", function() {
+    return this.get("store").peekAll("code");
+  }),
+
+  filteredPackageTypes: Ember.computed("allPackageTypes", function() {
+    const pkgTypes = this.get("allPackageTypes");
+    switch (this.get("storageType")) {
```
Darshan-Chauhan: this can be refactored as

```suggestion
    const pkgTypes = this.get("allPackageTypes");
    const storageType = this.get("storageType");
    return storageType ? pkgTypes.filterBy(`allow_${storageType.toLowerCase()}`, true) :  pkgTypes;
```

app/controllers/items/detail.js
```diff
@@ -132,6 +139,29 @@ export default GoodcityController.extend(
         .get("lastObject");
     }),
 
+    infoTabSelected: Ember.computed("tabName", function() {
+      return (
+        ["storage_detail", "storage_content", "info"].indexOf(
```
Darshan-Chauhan: Any reason for hard cording these values?

app/controllers/items/detail.js
```diff
@@ -132,6 +139,29 @@ export default GoodcityController.extend(
...
+          this.get("tabName")
+        ) > -1
+      );
+    }),
+
+    listTabSelected: Ember.computed("tabName", function() {
+      return ["storage_content", "info"].indexOf(this.get("tabName")) > -1;
```
Darshan-Chauhan: ^

app/controllers/items/detail.js
```diff
@@ -132,6 +139,29 @@ export default GoodcityController.extend(
...
+
+    storageTypeName: Ember.computed("item", function() {
+      let storageType = this.get("item.storageType");
+      if (storageType) {
+        return storageType.get("name");
+      }
+    }),
+
+    isBoxOrPallet: Ember.computed("item", function() {
+      return ["Box", "Pallet"].indexOf(this.get("storageTypeName")) > -1;
```
Darshan-Chauhan: ^

app/controllers/items/detail.js
```diff
@@ -132,6 +139,29 @@ export default GoodcityController.extend(
...
+      );
+    }),
+
+    listTabSelected: Ember.computed("tabName", function() {
+      return ["storage_content", "info"].indexOf(this.get("tabName")) > -1;
+    }),
+
+    storageTypeName: Ember.computed("item", function() {
+      let storageType = this.get("item.storageType");
+      if (storageType) {
```
Darshan-Chauhan:
```suggestion
      return storageType && storageType.get("name");
```

I think below suggestion will be a better approach for all scenarios like this in future.

https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining

so basically we can do

```suggestion
      return storageType?.get("name");
```

I think we should add support for optional chaining.

@patrixr , thoughts?

patrixr: I'm all for it, should be easy enough to add this plugin : )

app/controllers/items/detail.js
patrixr: Hardcoded values aside, I'm not sure I understand why the implementation isn't `this.get("tabName") === 'info'`
The info tab matches multiple tabs ? bit confused

abhinav098: I have done this purposefully, because I want to select the info tab, when inner tabs are also selected. 

In case of box and pallets, be have nested tabs.

So I want to select info tab, when any of the nested tab is selected.

cc- @abulsayyad123 @Darshan-Chauhan @patrixr

abhinav098: I have attached screen shot for reference.
Screen Shot 2019-12-26 at 7 11 34 PM

app/mixins/create_Inventory.js
```diff
@@ -0,0 +1,19 @@
+import Ember from "ember";
```
abulsayyad123: Why caps `I` for file name i.e `create_Inventory`?

abulsayyad123: There is no need to create this mixins. You can directly call this service from `.hbs`.
easiest way could be `{{action 'actionName' params target=packageService}}`

app/controllers/app_menu_list.js
```diff
@@ -1,44 +1,55 @@
...
+import Ember from "ember";
+import config from "stock/config/environment";
 const { getOwner } = Ember;
-import AjaxPromise from 'stock/utils/ajax-promise';
+import AjaxPromise from "stock/utils/ajax-promise";
+import createInventroyAction from "stock/mixins/create_Inventory";
+import storageType from "stock/mixins/storage-type";
 
-export default Ember.Controller.extend({
+export default Ember.Controller.extend(createInventroyAction, storageType, {
```
abulsayyad123: No need of importing these mixins if you are directly calling its action from handlebars.

app/templates/app_menu_list.hbs
```diff
@@ -30,11 +29,13 @@
...
         </li>
       {{/if}}
-      <li class="menu-list">
-        {{#link-to 'search_code' (query-params backToNewItem=false) class='menu-list-anchor inventory'}}
-          {{fa-icon 'plus-circle' size="lg"}}&nbsp;&nbsp;{{t "add_inventory_item"}}
-        {{/link-to}}
-      </li>
+      {{#each storageTypes as |storageType|}}
+        <li class="menu-list {{if storageType.disable 'disable'}}">
+          <a href="#" class="menu-list-anchor {{if storageType.disable 'disable'}}" {{action 'createInventory' storageType.name}} disabled={{storageType.disable}}>
```
abulsayyad123: As mentioned above, you could directly call that method from here.
`{{action 'createInventory' storageType.name target=packageService}}`

app/templates/items/detail/tabs/_item_detail.hbs
```diff
@@ -0,0 +1,179 @@
...
+</div>
+{{/toggle-item-options-list}}
+<div class="row detail-section">
+  {{#toggle-item-options-list item=item}}
+  <div class="columns small-12">
+    <div class="row">
+      <div class="columns small-3">{{t 'item_details.description'}} </div>
+      <div class="columns small-9 value description">
+        {{#form-control}}
+        {{item-description-textarea data-autoresize=true name="notes" value=item.notes required="required" pattern=".*\S.*" id=(concat 'desc' item.id) class="description-textarea" item=item}}
```
abulsayyad123: indentation is not proper.

app/templates/app_menu_list.hbs
abhinav098: sure, @abulsayyad123 i'll make that change. 
Thanks

config/environment.js
```diff
@@ -113,11 +114,11 @@ module.exports = function(environment) {
   };
 
   if (environment === "development") {
-    ENV.APP.API_HOST_URL = "http://localhost:3000";
+    ENV.APP.API_HOST_URL = "http://localhost:4000";
```
abulsayyad123: remove this

abhinav098: @abulsayyad123, added this purposefully so that we dont have to change it everytime.

app/components/goodcity/code-search-overlay.js
```diff
@@ -0,0 +1,130 @@
...
+    this.set("fetchMoreResult", true);
+  },
+
+  clearHiglight() {
+    Ember.$("em").replaceWith(function() {
+      return this.innerHTML;
+    });
+  },
+
+  highlight() {
```
abulsayyad123: If these highlighting logic repeated, find a way to make it more generic. Try to avoid using of `$` here.

PR #569, opened 2019-12-17
Redesignation from designate popup + GCW-2267 + GCW-2930

patrixr:
### Description

The previous designation changes came with a side effect : Loss of re-designation feature

This restores that. The main flow this impact is : 

Order detail page > Add item to order > Item list page > Click on item which is designated to another order

### Feature (reimplemented)

Allow to un-designate from a previous order from the designation popup.
In the case where a package is designated to multiple order (will not happen as long as stockit is alive), a link to the publishing tab of the item is provided where the user can undesignate as needed

### Bug Fixes 

- [GCW-2267](https://jira.crossroads.org.hk/browse/GCW-2267) (back navigation)
- [GCW-2930](https://jira.crossroads.org.hk/browse/GCW-2930) (back navigation)


app/templates/items/detail.hbs
```diff
@@ -1,13 +1,13 @@
 <section class="main-section item_details_screen">
   <div class="search">
     <div class="row profile_container">
-      {{#link-to "items.index" tagName='span' href=false}}
+      {{#back-link}}
```
abhinav098: I think this will create the item detail tabs back and forth issue just like order tabs. Because AFAIK, back-link is nothing but `window.history.back()`.

patrixr: There's always potential issues, but it's the behavior we want eventually. We'll fix the place it's navigating from later on if it does :)

PR #573, opened 2019-12-19
Gcw 2853 support printer selection in ui feature live

akshaygwd:
Ticket link 
https://jira.crossroads.org.hk/browse/GCW-2853

Feature: (LIVE pr for printer feature :stock)
 This ticket provides a UI for end users to select a printer and to set their own default printer when creating new packages

Please review: 
Thanks

### DIFF
--- a/app/controllers/items/new.js
+++ b/app/controllers/items/new.js
@@ -53,9 +53,10 @@ export default GoodcityController.extend(
     isAllowedToPublish: false,
     imageKeys: Ember.computed.localStorage(),
     i18n: Ember.inject.service(),
+    session: Ember.inject.service(),
     packageService: Ember.inject.service(),
+    printerService: Ember.inject.service(),
     cancelWarning: t("items.new.cancel_warning"),
-
     displayFields: Ember.computed("code", function() {
       let subform = this.get("code.subform");
       return this.returnDisplayFields(subform);
@@ -69,6 +70,20 @@ export default GoodcityController.extend(
     locale: function(str) {
       return this.get("i18n").t(str);
     },
+
+    allAvailablePrinters: Ember.computed(function() {
+      return this.get("printerService").allAvailablePrinters();
+    }),
+
+    selectedPrinterDisplay: Ember.computed("selectedPrinterId", function() {
+      const printerId = this.get("selectedPrinterId");
+      if (printerId) {
+        const printer = this.store.peekRecord("printer", printerId);
+        return { name: printer.get("name"), id: printer.id };
+      } else {
+        return this.get("allAvailablePrinters")[0];
+      }
+    }),
 
     setLocation: Ember.observer("scanLocationName", function() {
       var scanInput = this.get("scanLocationName");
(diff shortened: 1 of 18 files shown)


app/controllers/items/new.js
```diff
@@ -70,6 +71,20 @@ export default GoodcityController.extend(
...
+      return this.get("printerService").allAvailablePrinters();
+    }),
+
+    selectedPrinterDisplay: Ember.computed("selectedPrinterId", function() {
+      const printerId = this.get("selectedPrinterId");
+      if (printerId) {
+        const printer = this.store.peekRecord("printer", printerId);
+        return { name: printer.get("name"), id: printer.id };
+      } else {
+        return this.get("allAvailablePrinters")[0];
```
patrixr: A bit odd that `allAvailablePrinters` returns a list of objects `{name, id}`
This object structure is only needed in the page for the dropdown list (I assume) so you can build it here rather than in the service.

Otherwise, one day someone will call your service method, expecting a printer model, but instead will receive this object

app/routes/items/new.js
```diff
@@ -135,5 +152,6 @@ export default AuthorizeRoute.extend({
       }
       window.localStorage.setItem("isSelectLocationPreviousRoute", false);
     }
+    this.setupPrinterId(controller);
```
abulsayyad123: Just for future reference. No need to pass `controller` here. Controller is available throughout route file.

app/services/printer-service.js
```diff
@@ -0,0 +1,22 @@
...
+  session: Ember.inject.service(),
+
+  allAvailablePrinters() {
+    return this.get("store")
+      .peekAll("printer")
+      .sortBy("id")
+      .map(printer => printer.getProperties("name", "id"));
+  },
+  updateUserDefaultPrinter(printerId) {
+    new AjaxPromise(
```
abulsayyad123: Change it in future PR's. Inherit this file from `APIBase` service to avoid `AjaxPromise` method usage.

app/services/session.js
```diff
@@ -17,6 +17,10 @@ export default Ember.Service.extend({
     return store.peekAll("user_profile").get("firstObject") || null;
   }).volatile(),
 
+  loggedInUser: Ember.computed(function() {
```
patrixr:
```suggestion
  loggedInUser: Ember.computed("currentUser.id", function() {
```

app/models/user.js
```diff
@@ -13,10 +13,8 @@ export default Addressable.extend({
   lastConnected: attr("date"),
   lastDisconnected: attr("date"),
   i18n: Ember.inject.service(),
-
-  // permission:  belongsTo('permission', { async: false }),
```
patrixr: How come those are commented out ? 🤔

akshaygwd: these were already commented before so i removed those

app/services/session.js
akshaygwd: @patrixr This is done

app/controllers/items/new.js
akshaygwd: @patrixr i made the changes there is not need of creating a object i can use peekAll printer as well.

PR #577, opened 2019-12-31
GCW-2931-Request cancellation reason for orders

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2931

### What does this PR do?

FEATURE: Adding cancellation Reason for Orders. Also, removing cancellation reason on resubmitting the order.

BUG:  Fixed commented specs

### Impacted Areas

Go to specific order. Click "Cancel order" popup appears with cancellation message and options.

### Linked PR's:

https://github.com/crossroads/api.goodcity/pull/885


app/controllers/orders/detail.js
```diff
@@ -54,6 +57,25 @@ export default GoodcityController.extend({
...
+  }),
+
+  isOtherCancellationReason: Ember.computed("cancellationReasonId", function() {
+    const selectedCancelledReason = this.store.peekRecord(
+      "cancellation_reason",
+      this.get("cancellationReasonId")
+    );
+    return (
+      selectedCancelledReason.get("name") ===
+      this.get("i18n").t("order_cancellation_reason.other").string
```
patrixr: maybe better to use a constant for this instead of a translation string. Might fail in Chinese. And it'll certainly break if we ever want to change the UI text

app/controllers/orders/detail.js
```diff
@@ -159,6 +181,23 @@ export default GoodcityController.extend({
   },
 
   actions: {
+    cancelOrder() {
+      const params = {
```
patrixr: It'd be nice to let the service build the params object itself, the controller doesn't really need to.
Definitely not a big deal, just mentioning it

app/services/order-service.js
```diff
@@ -1,7 +1,16 @@
...
+
   ordersCountFor(userId) {
     return this.GET(`/users/${userId}/orders_count`);
+  },
+
+  changeOrderState(order, params) {
+    return this.PUT(`/orders/${order.id}/transition`, params).then(data => {
+      data["designation"] = data["order"];
+      this.get("store").pushPayload(data);
+    });
```
patrixr: Suggestion :

```javascript
changeOrderState(order, state, opts = {}) {
    return this.PUT(`/orders/${order.id}/transition`, { transition: state, ...opts }).then(data => {
      data["designation"] = data["order"]; // todo: consolidate the naming under 'order'
      this.get("store").pushPayload(data);
    });
},

// shorthand method to cancel orders
cancelOrder(order, reason) {
   return this.changeOrderState(order, 'cancel', { reason });
}
```

That allows you to call it like this : 

```
service.changeOrderState(order, 'cancel');
service.cancelOrder(order, 'pure evil');
```

Which is a bit more readable (imo).
This is not a change request, more for the sake of conversation :)

PR #581, opened 2020-01-06
GCW-2928-edit subform

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2928

### What does this PR do?

BUG:  
1) when changing the package type from one with extra fields that have content in them to the same package type (or a different one with the same extra fields), the contents in the 'Country of Origin' dropdown field is deleted.
2) older items with a package type that should have the extra fields (but don't) can have their package type changed to a different package type and back again to have the extra fields, but a different package type must be first selected. If the the same package type is first chosen and then a different package type, the confirmation pop-up appears but doesn't continue past the package type selection page.

### Impacted Areas

Item detail screen.

### DIFF
--- a/app/controllers/items/select_code.js
+++ b/app/controllers/items/select_code.js
@@ -72,7 +72,10 @@ export default SearchCode.extend({
     const packageParams = {
       package_type_id: type.get("id")
     };
-    if (!this.isSamePackage(type)) {
+    if (
+      !this.isSamePackage(type) ||
+      (!item.get("detailId") && this.isSubformPackage(type))
+    ) {
       packageParams.detail_type = capitalize(type.get("subform"));
     }
     if (deleteDetailId) {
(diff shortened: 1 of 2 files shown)


app/controllers/items/select_code.js
```diff
@@ -23,9 +23,11 @@ export default SearchCode.extend({
     const item = this.get("item");
     const type = item.get("detailType");
     const detailId = item.get("detailId");
-    await this.runTask(
-      this.get("subformDetailService").deleteDetailType(type, detailId)
-    );
+    if (type) {
```
abhinav098: @abulsayyad123 , its better to check type one of ["electrical", "computer", "computer_accessory"] instead of just checking type because some of the package_types/items have detail_type as "medical" and it might cause an issue going forward. 

:)

abulsayyad123: This check is for delete subform if there exist subform.

abhinav098: okay, makes sense 👍

PR #587, opened 2020-01-08
GCW-2859 adding/removing items from box

abhinav098:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2859

### What does this PR do?

FEATURE: Allows user to add/remove items to box/pallet

### Mockup Link

https://xd.adobe.com/view/224f7314-f1c0-40b6-7d7a-929e616be6bb-547e/screen/b329497c-a23f-4063-89c8-9ae7b71d253a/Main-Menu?x_product=xd-jiraserver/1.0

### Linked PR's:
https://github.com/crossroads/api.goodcity/pull/891


app/components/goodcity/add-item-overlay.js
```diff
@@ -0,0 +1,99 @@
...
+      .catch(response => {
+        this.get("messageBox").alert(
+          (response.responseJSON && response.responseJSON.errors[0]) ||
+            this.get("i18n").t("unexpected_error")
+        );
+      })
+      .finally(() => {
+        loadingView.destroy();
+      });
+  },
```
patrixr: Suggestion: using an async function with `mixins/async`

```suggestion
  addRemoveItem(params) {
    return this.runTask(async () => {
      const data = await this.get("packageService").addRemoveItem(this.get("entity.id"), params);

      this.get("store").pushPayload(data);
      this.set("open", false);
      this.sendAction("onConfirm");
    }, ERROR_STRATEGIES.MODAL)
  },
```

app/components/goodcity/locations-search-overlay.js
```diff
@@ -43,7 +44,7 @@ export default SingletonComponent.extend(SearchMixin, {
 
     async selectLocation(location) {
       const callback = this.getWithDefault("onSelect", _.noop);
-
+      this.set("selectedLocation", location);
```
patrixr: We should **not** store a location in that component. It should only **return** it.
Whoever calls the component (via the service) can store the returned location and do whatever it needs with it.

This is a stateful red flag

Components should be designed in a generic manner, and should not bend to the desires of the controllers :)

app/components/goodcity/add-item-overlay.js
```diff
@@ -0,0 +1,99 @@
+import Ember from "ember";
+const { getOwner } = Ember;
+import SingletonComponent from "../base/global";
+import _ from "lodash";
+
+export default SingletonComponent.extend({
```
patrixr: No need to inherit from `SingletonComponent`, that was made only for the global ones. You can inherit from `Ember.Component` since they are used within the pages and therefore get destroyed alongside it

patrixr: Applies to other components too : )

app/controllers/items/detail.js
```diff
@@ -254,6 +264,54 @@ export default GoodcityController.extend(
...
+            this.store.pushPayload(data);
+            this.set("associatedPackages", data.items);
+          })
+          .finally(() => {
+            this.hideLoadingSpinner();
+          });
+      },
+
+      openLocationSearch(item) {
+        this.set("locationService.openLocationSearch", true);
```
patrixr: The way to use the location service is : 

```javascript
const selectedLocation = await this.get('locationService').userPickLocation( /* opts */ );
// do something with the selected location
```
We shouldn't store anything in the component, or modify the service's properties directly. That's a dangerous game

app/models/item.js
```diff
@@ -64,6 +85,21 @@ export default cloudinaryUrl.extend({
...
+    switch (this.get("storageTypeName")) {
+      case "Box":
+        return "box-open";
+      case "Pallet":
+        return "pallet";
+      default:
+        return "tag";
+    }
+  }),
+
```
patrixr: We've used this pattern in the past, but the more I see it the more I feel it shouldn't be in the model.
I think a helper is what we should use

```javascript
// storage-icon.js
export default Ember.Helper.helper(function([storageType]) {
  return {
    box: "box-open",
    pallet: "pallet"
  }[storageType.toLowercase()] || "tag"
});
```

and in the `.hbs`

```handlebars
{{fa-icon (storage-icon record.storageType)}
```

Models _should_ be data focused. I've made that mistake myself in the past, I think we should start turning things around

app/models/item.js
```diff
@@ -40,22 +40,43 @@ export default cloudinaryUrl.extend({
...
+  }),
 
   storageTypeId: attr("number"),
   storageType: belongsTo("storage_type", {
     async: true
   }),
 
-  ordersPackages: hasMany("ordersPackages", { async: true }),
+  addedQuantity: attr("number"),
+  inHandQuantity: attr("number"),
```
patrixr: If you could rename to `onHandQuantity`, both to be more correct and to match the upcoming counter caches column that we'll be adding : )

app/services/location-service.js
```diff
@@ -19,6 +19,7 @@ export default ApiBaseService.extend(NavigationAwareness, {
   init() {
     this._super(...arguments);
     this.set("openLocationSearch", false);
+    this.set("selectedLocation", null);
```
patrixr: To remove, as mentioned in other comments

app/components/goodcity/add-item-overlay.js
patrixr: Other note: It cannot be **both** bound to a service **and** within a page's`.hbs` file.

That will break, since the page gets destroyed but the service remains.

Options:

* 1) Component is inside a page's template, and is bound to the controller _(normal kind)_  
* 2) Component is inside the `_global_components.hbs` template and is bound to a service

app/templates/_global_components.hbs
```diff
@@ -12,6 +12,7 @@
   {{goodcity/locations-search-overlay
     open=s.openLocationSearch
     onSelect=s.onLocationSelected
+    selectedLocation=s.selectedLocation
```
patrixr: todo: remove

app/mixins/search_resource.js
```diff
@@ -122,10 +122,11 @@ export default Ember.Mixin.create({
    * @memberof SearchMixin
    * @returns {object} the search parameters
    **/
-  getSearchQuery() {
+  getSearchQuery(filter = undefined) {
     return {
       searchText: this.get("searchText"),
-      ...this.get("searchProps")
+      ...this.get("searchProps"),
+      filter_box_pallet: filter
```
patrixr: `search_resource` is a generic reusable mixin. It **cannot** have any references to box/pallets.

I suggest you extend the method in the component/controller which includes the mixin

abhinav098: sure @patrixr , will make this change.

app/components/goodcity/add-item-overlay.js
abhinav098: 👍

app/components/goodcity/locations-search-overlay.js
abhinav098: will check, and fix this @patrixr .

app/components/goodcity/add-item-overlay.js
```diff
@@ -0,0 +1,84 @@
+import Ember from "ember";
+import AsyncMixin, { ERROR_STRATEGIES } from "../../mixins/async";
```
swatijadhav: use absolute file path.

app/components/goodcity/add-item-overlay.js
```diff
@@ -0,0 +1,84 @@
...
+      return this.get("pkg").get("packagesLocations");
+    }
+  },
+
+  calculateSumFor(parameter) {
+    let quantities = [];
+    if (this.get("pkg.packagesLocations")) {
+      this.get("pkg.packagesLocations").map(value => {
+        quantities.push(value.get(parameter));
+      });
```
swatijadhav:
```js
quantities = this.get("pkg.packagesLocations").map(value => value.get(parameter) );
```

app/components/goodcity/add-item-overlay.js
```diff
@@ -0,0 +1,84 @@
...
+    }, ERROR_STRATEGIES.MODAL);
+  },
+
+  actions: {
+    moveItemToBox() {
+      let pkg = this.get("pkg");
+      if (pkg) {
+        this.get("pkgLocations").forEach(pkgLocation => {
+          let selectedQuantity = pkgLocation.get("defaultQuantity");
+          if (pkgLocation.get("hasDirtyAttributes") && selectedQuantity != 0) {
```
swatijadhav:
```suggestion
          if (pkgLocation.get("hasDirtyAttributes") && selectedQuantity !== 0) {
```

app/components/goodcity/add-item-overlay.js
```diff
@@ -0,0 +1,85 @@
...
+    let quantities = [];
+    if (this.get("pkg.packagesLocations")) {
+      quantities = this.get("pkg.packagesLocations").map(value =>
+        value.get(parameter)
+      );
+    }
+
+    return quantities.reduce((accumulator, value) => {
+      return accumulator + (!(value == "") && parseInt(value));
+    }, 0);
```
swatijadhav:
```js
return quantities.reduce((accumulator, value) => accumulator + parseInt(value || 0), 0);
```

app/components/goodcity/add-item-overlay.js
```diff
@@ -0,0 +1,85 @@
...
+    }, ERROR_STRATEGIES.MODAL);
+  },
+
+  actions: {
+    moveItemToBox() {
+      let pkg = this.get("pkg");
+      if (pkg) {
+        this.get("pkgLocations").forEach(pkgLocation => {
+          let selectedQuantity = pkgLocation.get("defaultQuantity");
+          if (pkgLocation.get("hasDirtyAttributes") && selectedQuantity !== 0) {
```
swatijadhav:
```suggestion
          if (pkgLocation.get("hasDirtyAttributes") && selectedQuantity) {
```

abhinav098: @swatijadhav this is intentionally done, i want to proceed only if selectedQuantity is not 0. `selectedQuantity !== 0`.

abhinav098: ohh, understood, `if(true && 0)` will go to else. 
Thanks @swatijadhav

app/components/goodcity/item-search-overlay.js
```diff
@@ -0,0 +1,54 @@
...
+            associated_package_types: this.get("associatedPackageTypes"),
+            withInventoryNumber: true,
+            filter_box_pallet: true
+          }
+        )
+      );
+      let items = await this.get("store").query("item", params);
+      return items.filter(function(item) {
+        return item.get("onHandQuantity") > 0;
+      });
```
swatijadhav:
```
return items.filter(item => item.get("onHandQuantity"));
```

app/components/message-box.js
```diff
@@ -1,16 +1,14 @@
-import Ember from 'ember';
-import messageBox from '../templates/components/message-box';
+import Ember from "ember";
+import messageBox from "../templates/components/message-box";
```
swatijadhav: Use absolute path.

app/styles/templates/items/_detail.scss
```diff
@@ -238,14 +235,64 @@
...
+          height: 2rem;
+          line-height: 1.2rem;
+          left: 0.65rem;
+          bottom: -4%;
+        }
+      }
+
+      @media #{$medium-only} {
+        // left: 0.5rem;
+      }
```
swatijadhav: This style can be removed.

app/templates/components/goodcity/add-item-overlay.hbs
```diff
@@ -0,0 +1,95 @@
...
+            <div class="content item-img">
+              <img src={{pkg.thumbImageUrl}} class="thumb" />
+            </div>
+            <div class="content item-description">
+              <div>
+                {{pkg.notes}}
+              </div>
+              <div>
+                {{pkg.notes}}
+              </div>
```
swatijadhav: Are we intentionally showing `notes` text twice?

abhinav098: It is added as per the mockup @swatijadhav .

app/templates/items/_item_description.hbs
```diff
@@ -1,8 +1,8 @@
 {{#toggle-item-options-list}}
   <div class="small-6 large-8 columns name_details">
-    <div class="ellipsis name one-line-ellipsis" style="display: inline!important;">{{item.quantity}} x {{item.code.name}}</div>
+    <div class="ellipsis name one-line-ellipsis" style="display: inline!important;">{{item.onHandQuantity}} x {{item.code.name}}</div>
     <span class="available-designate-dispatch-count" style="margin-left: 0.25rem;">
-      {{item.quantity}}/{{item.totalDesignatedQty}}/{{item.totalDispatchedQty}}
+      {{item.onHandQuantity}}/{{item.totalDesignatedQty}}/{{item.totalDispatchedQty}}
```
abulsayyad123: 👍

app/templates/items/detail/tabs/storage_content.hbs
```diff
@@ -1,7 +1,43 @@
 {{partial 'items/detail/tabs/_sub_tabs' }}
-
 <div class="row">
-  <div class="small-12 columns add-item" align="center">
-    <button class="button expand" disabled={{disableBoxPalletItemAddition}}>Add item to {{storageTypeName}}</button>
+  <div class="large-12 medium-9 small-9 columns add-item" align="center">
+    <button class="button expand" {{action 'openItemsSearch' item target=packageService}}
```
patrixr: The overlay is within the page, but the service has an `openItemsSearch` method. Calling it on any other page won't work. It should probably go to the controller.

abhinav098: I'll do that change @patrixr .
Thanks. 
Please review API PR=> https://github.com/crossroads/api.goodcity/pull/891 as well.

PR #588, opened 2020-01-10
GCW-2932-infinite scroll for order active tab

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2932

### What does this PR do?

FEATURE:  This PR adds infinite scroll to orders active items page. 
Till now we were making separate request on loading packages associated to orders. This PR removes this extra requests.

### Impacted Areas
Goto any specific Order >  Goto Active Items pages.


app/routes/orders/active_items.js
```diff
@@ -1,3 +1,8 @@
 import detail from "./detail";
 
-export default detail.extend({});
+export default detail.extend({
+  async setupController(controller, model = {}) {
+    await this._super(controller, model);
+    controller.on();
```
patrixr: If you're calling `on(), I'm guessing you should probably call `off()` when leaving the page :)

app/controllers/orders/detail.js
```diff
@@ -159,6 +133,18 @@ export default GoodcityController.extend({
...
+    loadOrdersPackages(pageNo) {
+      const params = this.trimQuery(
+        _.merge(
+          { order_id: this.get("orderId") },
+          this.getPaginationQuery(pageNo)
+        )
+      );
+      return this.get("store")
+        .query("orders_package", params)
+        .then(results => results);
```
patrixr: this `then` statement doesn't seem needed since you're just returning the value straight away

app/controllers/orders/detail.js
```diff
@@ -2,14 +2,21 @@ import Ember from "ember";
...
 import AjaxPromise from "stock/utils/ajax-promise";
 import GoodcityController from "../goodcity_controller";
+import SearchMixin from "stock/mixins/search_resource";
 import _ from "lodash";
 const { getOwner } = Ember;
 
-export default GoodcityController.extend({
+export default GoodcityController.extend(SearchMixin, {
+  autoLoad: true,
+  /**
```
patrixr: I've added JSDoc to the project.

So commenting something with `/**` as a start will get captured in the doc, but the documentation tool won't know to which module it belong.

So options are:
* Document the controller itself (there are examples in the project)
* Simply start the comment using `/*` or `//`

Note: Documenting Controllers, although not a bad thing, is definitely not the most useful thing to document generally, since they're not meant to be re-used

app/models/orders_package.js
```diff
@@ -13,7 +13,7 @@ export default Model.extend({
   state: attr("string"),
   allowedActions: attr(),
 
-  item: belongsTo("item", { async: true }),
+  item: belongsTo("item", { async: false }),
```
namrataukirde: :+1:

PR #594, opened 2020-01-21
added overflow auto to infinite container

akshaygwd:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2978

### What does this PR do?

Bug:
This PR removes the unwanted scroll appears of internet explorer on the orders page even if list elements are less

before:
Untitled (1)

after:
Untitled2
Please review

### DIFF
--- a/app/styles/templates/orders/_index.scss
+++ b/app/styles/templates/orders/_index.scss
@@ -8,26 +8,25 @@ $header-height: 3em;
     padding-left: 0;
     padding-right: 0;
 
-
     .order_process_summary .value {
-      >span {
+      > span {
         margin-right: 0.7em;
       }
     }
-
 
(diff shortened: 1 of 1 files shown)


app/styles/templates/orders/_index.scss
```diff
@@ -112,7 +110,9 @@ $header-height: 3em;
     }
   }
 
-
+  .infinite-list-container {
+    overflow-y: auto !important;
```
namrataukirde: I will suggest to check CSS on infinity-scroll plugin and decide it this really needed? As infinity-scroll is our own plugin.

akshaygwd: Okay got it

PR #595, opened 2020-01-29
Dispatch per location

patrixr:
### New dispatch flow

* User clicks on the `dispatch` action
* User selects a location to dispatch from _(skipped if only one location)_
* User selects the quantity to dispatch _(locked to full quantity of that location until stockit is dead)_
* User triggers the dispatch
* The API applies the action, as seen here : https://github.com/crossroads/api.goodcity/pull/904

### Screenshot

Screenshot 2020-01-29 at 4 12 39 PM


app/helpers/clamp.js
```diff
@@ -0,0 +1,33 @@
...
+  return function(input) {
+    let num;
+
+    if (input instanceof jQuery.Event) {
+      num = Number(input.target.value);
+    } else {
+      num = Number(input);
+    }
+
+    if (num < min) {
```
abulsayyad123: is num = min or max case possible ?

app/serializers/application.js
```diff
@@ -1,3 +1,29 @@
...
+import _ from "lodash";
 
-export default ActiveModelSerializer.extend();
+// @TODO: rename the item model into package and remove those aliases
+const ALIASES = {
+  package: "item",
+  packages: "items"
+};
+
+function normalize(payload) {
```
abulsayyad123: 👍

app/helpers/clamp.js
patrixr: Those are valid cases, it's an inclusive range. So no need to do anything in that case

app/serializers/application.js
patrixr: @abulsayyad123 I think we can remove the normalization in the `orders_package` serializer, since it inherits from here. This applies to everything. Haven't done it for now but thinking about it

abulsayyad123: i had the same question regarding `orders_packages` serializer. Can we remove it ? I thought it was part of separate entity, so application serializer wouldn't change it.

patrixr: I believe we can : )

PR #603, opened 2020-02-12
Fix for GCW-2904 (part 2)

patrixr:
More backlink fixes for ticket GCW-2904:

Change: Switch from `transitionTo` to `history.back()`

### DIFF
--- a/app/components/cancel-order.js
+++ b/app/components/cancel-order.js
@@ -1,9 +1,9 @@
 import Ember from "ember";
-import AjaxPromise from 'stock/utils/ajax-promise';
+import AjaxPromise from "stock/utils/ajax-promise";
 const { getOwner } = Ember;
 
 export default Ember.Component.extend({
-  tagName: 'a',
+  tagName: "a",
   isCancel: true,
   messageBox: Ember.inject.service(),
   store: Ember.inject.service(),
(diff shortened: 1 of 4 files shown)


app/components/cancel-order.js
```diff
@@ -1,9 +1,9 @@
 import Ember from "ember";
-import AjaxPromise from 'stock/utils/ajax-promise';
+import AjaxPromise from "stock/utils/ajax-promise";
 const { getOwner } = Ember;
 
 export default Ember.Component.extend({
-  tagName: 'a',
+  tagName: "a",
```
steveyken: Mission critical change. :100:

app/helpers/back-action.js
```diff
@@ -23,7 +23,7 @@ export default Ember.Helper.extend({
 
       const noHistory = window.history.length === 1;
       const deepLinked =
-        document.referrer && document.referrer.indexOf(config.APP.NAME) < 0;
+        document.referrer && document.referrer.indexOf(config.APP.ORIGIN) < 0;
```
abulsayyad123: What does `document.referrer` returns ?

patrixr: It's just a string with the url of the page that redirected to your app.

So if on "jira.com" I click on a `stock.goodcity.hk` link, then referrer will be Jira

app/components/cancel-order.js
patrixr: That's the prettifier hook :'(

app/helpers/back-action.js
patrixr: so if the referrer is Stock, that means I navigated from my app -> to my app

abulsayyad123: Thanks 👍  for clarifying.

PR #607, opened 2020-02-17
Scanning Barcode issue coming for the second instance is  resolved

shreyas098:
### What does this PR do?
Scanning the same barcode twice or more makes it viewable

BUG: Scanning the same barcode twice makes inventory no. unviewable 

NOTE: The observer present was triggering only once based based on query params for which another observer was added to have a check on search Text which resolved the issue.

### main change : scannedText

### DIFF
--- a/app/controllers/items/index.js
+++ b/app/controllers/items/index.js
@@ -33,6 +33,18 @@ export default Ember.Controller.extend(SearchMixin, {
     }
   }),
 
+  scannedText: Ember.observer("searchText", function() {
+    const searchInput = this.get("searchText") || "";
+    const sanitizeString = this.sanitizeString(searchInput);
+    //if (sanitizeString) {
+    this.set("searchInput", sanitizeString);
+    //}
+  }),
+
+  hasSearchText: Ember.computed("searchText", function() {
+    return Ember.$.trim(this.get("searchText")).length;
+  }),
+
   getFilterQuery() {
     let filterService = this.get("filterService");
     let utilities = this.get("utilityMethods");
@@ -74,6 +86,9 @@ export default Ember.Controller.extend(SearchMixin, {
         .then(results => {
           return results;
         });
+    },
+    clearSearch() {
+      this.set("searchText", "");
     }
   }
 });
--- a/app/controllers/orders/detail.js
+++ b/app/controllers/orders/detail.js
@@ -37,9 +37,15 @@ export default GoodcityController.extend(SearchMixin, {
       .get("lastObject");
   }),
 
-  highlightSelectedTab: Ember.computed("tabName", function() {
+  highlightSelectedTabSummary: Ember.computed("tabName", function() {
     return (
       ["client_summary", "contact_summary"].indexOf(this.get("tabName")) >= 0
+    );
+  }),
+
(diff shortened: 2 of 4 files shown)


app/controllers/items/index.js
```diff
@@ -33,6 +33,18 @@ export default Ember.Controller.extend(SearchMixin, {
     }
   }),
 
+  scannedText: Ember.observer("searchText", function() {
+    const searchInput = this.get("searchText") || "";
+    const sanitizeString = this.sanitizeString(searchInput);
+    //if (sanitizeString) {
```
abhinav098: @shreyas098 , why is this commented? If this is not required please remove it.

Keeping a check and then setting the `searchInput` is better.

Please let me know your thoughts.

app/controllers/orders/detail.js
```diff
@@ -37,12 +37,18 @@ export default GoodcityController.extend(SearchMixin, {
...
     return (
       ["client_summary", "contact_summary"].indexOf(this.get("tabName")) >= 0
     );
   }),
 
+  highlightSelectedTabGoods: Ember.computed("tabName", function() {
+    return (
+      ["active_items", "requested_items"].indexOf(this.get("tabName")) >= 0
+    );
+  }),
```
abhinav098: As far as I know, this fix is for a separate ticket, please create a new branch from master and move this code to that branch @shreyas098 . We should not merge two tickets together while fixing issues or any other thing for that matter.

app/templates/orders/_order_tabs.hbs
```diff
@@ -3,14 +3,15 @@
     <div class="small-12 columns tab_row">
       <dl class="tabs" data-tab>
         <dd class="small-3 medium-3 columns text-center
-          {{if highlightSelectedTab 'active'}}">
+          {{if highlightSelectedTabSummary 'active'}}">
```
abhinav098: same as above.

app/templates/orders/_order_tabs.hbs
```diff
@@ -3,14 +3,15 @@
...
           {{#link-to 'orders.contact_summary' model.id replace=true}}
           <i>{{fa-icon 'id-card'}}</i>
           <div>{{t "order_tabs.summary"}}</div>
           {{/link-to}}
         </dd>
 
-        <dd class="small-3 medium-3 columns text-center {{if (is-equal tabName 'active_items') 'active'}}">
+        <dd class="small-3 medium-3 columns text-center
+        {{if highlightSelectedTabGoods 'active'}}">
```
abhinav098: same as above.

app/controllers/orders/detail.js
```diff
@@ -37,12 +37,18 @@ export default GoodcityController.extend(SearchMixin, {
       .get("lastObject");
   }),
 
-  highlightSelectedTab: Ember.computed("tabName", function() {
+  highlightSelectedTabSummary: Ember.computed("tabName", function() {
```
abhinav098: Kindly move these fixes to a separate branch.

shreyas098:
Two Branches were merged mistakenly so creating new pull request for the same individual branches

PR #608, opened 2020-02-17
Gcw 3039 barcode scanner on new items creation

shreyas098:
### What does this PR do?
Opens the barcode and doesn't crashes at new items creation

BUG:barcode scanner on new item creation breaks.

NOTE: It had problem with the **this** structure, which was resolved

### main change :new.js


shreyas098:
Two Branches were merged mistakenly so creating new pull request for the same individual branches

PR #612, opened 2020-02-18
GCW-2101 Scanner gets the item's number on search tab if scanned more than one…

shreyas098:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-2101

### What does this PR do?

Scanning the same barcode twice or more makes it viewable

BUG: Scanning the same barcode twice makes inventory no. unviewable

NOTE: The observer present was triggering only once based based on query params for which another observer was added to have a check on search Text which resolved the issue.

###main change :
 scannedText

### DIFF
--- a/app/controllers/items/index.js
+++ b/app/controllers/items/index.js
@@ -33,6 +33,15 @@ export default Ember.Controller.extend(SearchMixin, {
     }
   }),
 
+  scannedText: Ember.observer("searchText", function() {
+    const searchInput = this.get("searchText") || "";
+    this.set("searchInput", this.sanitizeString(searchInput));
+  }),
+
+  hasSearchText: Ember.computed("searchText", function() {
+    return Ember.$.trim(this.get("searchText")).length;
+  }),
+
   getFilterQuery() {
     let filterService = this.get("filterService");
     let utilities = this.get("utilityMethods");
@@ -74,6 +83,10 @@ export default Ember.Controller.extend(SearchMixin, {
         .then(results => {
           return results;
         });
+    },
+
+    clearSearch() {
+      this.set("searchText", "");
     }
   }
 });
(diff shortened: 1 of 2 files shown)


app/controllers/items/index.js
```diff
@@ -74,6 +83,10 @@ export default Ember.Controller.extend(SearchMixin, {
         .then(results => {
           return results;
         });
+    },
+
+    clearSearch() {
```
abhinav098: @shreyas098 where are you using this action ?

app/controllers/orders/index.js
```diff
@@ -70,6 +74,10 @@ export default Ember.Controller.extend(SearchMixin, {
           this.afterSearch(results);
           return results;
         });
+    },
+
+    clearSearch() {
```
abhinav098: same question for this action too.

app/controllers/orders/index.js
```diff
@@ -25,6 +25,10 @@ export default Ember.Controller.extend(SearchMixin, {
     }
   },
 
+  hasSearchText: Ember.computed("searchText", function() {
```
abhinav098: This property is also newly added. Where are you using it ?
Did you forget to commit the template file ?

shreyas098: So the action is already present hbs file but the property and actions were not present in respective js file,so added that property and action for the same

app/controllers/items/index.js
shreyas098: For the cross button which appears on the search tab

app/controllers/orders/index.js
shreyas098: For the cross button which appears on the search tab

app/controllers/orders/index.js
abhinav098: ohh i see. Thanks for clarifying.

abhinav098:
@shreyas098 , I have approved this pull request. 
Please check the code climate issues and fix them if possible.
https://codeclimate.com/github/crossroads/stock.goodcity/pull/612
Thanks

PR #619, opened 2020-02-20
GCW-2859 bug fixes and suggestions implementation

abhinav098:
### Ticket Link: 

GCW-2859 https://jira.crossroads.org.hk/browse/GCW-2859

### What does this PR do?
This PR fixes the following issues:
1. Box should not be listed in item search while adding for a box
2. If item has qty 1, on click should add directly instead of showing the move overlay
3. Fix for: When an item or box is added to a box or pallet, error messages continually appear on the item or box page until the page is reloaded. ( async: true for packages locations was causing the issue )

### DIFF
--- a/app/components/goodcity/add-item-overlay.js
+++ b/app/components/goodcity/add-item-overlay.js
@@ -1,8 +1,7 @@
 import Ember from "ember";
-import AsyncMixin, { ERROR_STRATEGIES } from "stock/mixins/async";
 import _ from "lodash";
 
-export default Ember.Component.extend(AsyncMixin, {
+export default Ember.Component.extend({
   packageService: Ember.inject.service(),
   store: Ember.inject.service(),
   messageBox: Ember.inject.service(),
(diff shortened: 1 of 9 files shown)


app/components/goodcity/add-item-overlay.js
```diff
@@ -29,44 +28,72 @@ export default Ember.Component.extend(AsyncMixin, {
...
-        .then(this.set("open", false))
-        .catch(response => {
-          this.get("messageBox").alert(
-            response.responseText.errors[0] ||
-              this.get("i18n").t("unexpected_error")
-          );
-        })
-    );
+  resolvePromise() {
+    let promises = this.createPromises();
```
abulsayyad123: `createPromises` seems generic name. But this method name is not generic it is returning array of promises for Removing of Item.

app/components/goodcity/item-search-overlay.js
```diff
@@ -7,32 +7,66 @@ export default Ember.Component.extend(SearchMixin, {
...
   actions: {
     cancel() {
       this.set("searchText", "");
       this.set("open", false);
     },
 
     selectItem(item) {
-      this.set("open", false);
+      if (item.get("onHandQuantity") === 1) {
+        return this.addSingleton(item);
```
abulsayyad123: Why `return` ?

abhinav098: good catch, return not required, will remove it.

app/components/goodcity/add-item-overlay.js
abhinav098: sure 👍

app/components/goodcity/item-search-overlay.js
```diff
@@ -7,32 +7,66 @@ export default Ember.Component.extend(SearchMixin, {
...
       this.get("onConfirm")(item);
+      this.set("open", false);
     },
 
     clearSearch() {
       this.set("searchText", "");
     },
 
     async loadMoreItems(pageNo) {
+      let storageTypeName = this.get("storageTypeName");
```
abulsayyad123: If not used anywhere else you can directly pass it.

app/components/goodcity/item-search-overlay.js
abhinav098: @abulsayyad123 , return required because I don't want to proceed further if I get inside this block.

app/components/goodcity/add-item-overlay.js
```diff
@@ -29,44 +28,72 @@ export default Ember.Component.extend(AsyncMixin, {
...
+          quantity: selectedQuantity
+        };
+        promises.push(
+          this.get("packageService").addRemoveItem(
+            this.get("entity.id"),
+            params
+          )
+        );
+      }
+      pkgLocation.rollbackAttributes();
```
abulsayyad123: Not sure if `rollbackAttributes` asynchronous ?

app/components/goodcity/add-item-overlay.js
```diff
@@ -29,44 +28,72 @@ export default Ember.Component.extend(AsyncMixin, {
...
+        this.get("messageBox").alert(
+          error_message || this.get("i18n").t("unexpected_error")
+        );
+      });
+  },
+
+  isAddedQuantityInvalid() {
+    let pkgLocations = this.get("pkg.packagesLocations");
+    if (pkgLocations) {
+      return pkgLocations.getEach("isDefaultQuantityValid").includes(false);
```
namrataukirde: Can you try filterBy over here and let me know if that works? 🤔

app/components/goodcity/add-item-overlay.js
```diff
@@ -29,44 +28,72 @@ export default Ember.Component.extend(AsyncMixin, {
...
+      .catch(response => {
+        let error_message =
+          response.responseJSON && response.responseJSON.errors[0];
+        this.get("messageBox").alert(
+          error_message || this.get("i18n").t("unexpected_error")
+        );
+      });
+  },
+
+  isAddedQuantityInvalid() {
```
swatijadhav: Can we better rename it to `hasInvalidQuantity` ?

PR #661, opened 2020-03-06
GCW-3080-Scanner button as well as X button is working now

shreyas098:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3080

### What does this PR do?
BUG: Scanner is working fine in orders/items page and the issue of x button is fixed 
Screen Shot 2020-03-06 at 5 15 30 PM
Screen Shot 2020-03-06 at 5 15 59 PM

### DIFF
--- a/app/controllers/orders/items.js
+++ b/app/controllers/orders/items.js
@@ -24,15 +24,15 @@ export default Ember.Controller.extend(SearchMixin, {
 
   scannedItem: Ember.observer("searchInput", function() {
     const searchInput = this.get("searchInput") || "";
-    const sanitizeString = this.sanitizeString(searchInput);
-    if (sanitizeString) {
+    const sanitizedString = this.sanitizeString(searchInput);
+    if (sanitizedString) {
       this.set("searchText", sanitizeString);
     }
   }),
 
   scannedText: Ember.observer("searchText", function() {
-    const searchInput = this.get("searchText") || "";
-    this.set("searchInput", this.sanitizeString(searchInput));
+    const searchValue = this.get("searchText") || "";
+    this.set("searchInput", this.sanitizeString(searchValue));
   }),
 
   triggerDisplayDesignateOverlay() {
--- a/app/controllers/orders/items.js
+++ b/app/controllers/orders/items.js
@@ -22,6 +22,19 @@ export default Ember.Controller.extend(SearchMixin, {
     };
   },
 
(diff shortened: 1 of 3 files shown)


app/controllers/orders/items.js
```diff
@@ -22,6 +22,19 @@ export default Ember.Controller.extend(SearchMixin, {
     };
   },
 
+  scannedItem: Ember.observer("searchInput", function() {
```
patrixr: I'd recommend naming `observers` something like `scannedItemWatcher` as opposed to `scannedItem`

Otherwise it can very quickly become confusing, or conflict with actual properties of the controller : )

app/controllers/orders/items.js
```diff
@@ -22,6 +22,19 @@ export default Ember.Controller.extend(SearchMixin, {
...
+    const searchInput = this.get("searchInput") || "";
+    const sanitizedString = this.sanitizeString(searchInput);
+    if (sanitizedString) {
+      this.set("searchText", sanitizeString);
+    }
+  }),
+
+  scannedText: Ember.observer("searchText", function() {
+    const searchValue = this.get("searchText") || "";
+    this.set("searchInput", this.sanitizeString(searchValue));
```
patrixr: I notice a dangerous loop here, observers can trigger each others in a _seemingly_ infinite loop. Even though it might work for reasons X or Y, that is not a pattern we want to see in the app

app/controllers/orders/items.js
shreyas098: Yes will do that @patrixr

app/controllers/orders/items.js
shreyas098: Observers only trigger if they observe a change in value , so the flow goes as when the searchInput is made set for the first time inside scannedItem then scannedText will be called due to change in search text which at the end is setting searchInput, but this time scannedItem observer wont be triggered and the loop will break because there is no change in the searchInput value. So there is no possibility of any test cases running in infinite loop . @patrixr

patrixr: but there actually is a change to the searchinput : `this.set("searchInput", ...)`

I understand that there's no actual infinite loop, you would have noticed that. It is prevented by X or Y. But writing code that describes that pattern is a big code smell.

Or simply the fact that we have `this.sanitizeString(searchInput)` is weird.

All of the above are indications that there is a simpler and cleaner way to achieve this.

PR #672, opened 2020-03-18
GCW-3092-Add cache search filter results

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3092

### What does this PR do?
Added caching for search results on orders page. For now, I have used it only on `orders` index i.e search page. After testing , will apply on items search page as well.


app/utils/mem-cache.js
```diff
@@ -0,0 +1,38 @@
...
+  set(key, value) {
+    this.cache[key] = {
+      value: value,
+      expiry: Date.now() + this.timeout
+    };
+  }
+
+  has(key) {
+    return this.cache[key];
+  }
```
patrixr:
```suggestion
  has(key) {
    return !!this.get(key);
  }
```

app/mixins/search_resource.js
```diff
@@ -43,6 +43,7 @@ export default Ember.Mixin.create({
   onSearchTextChange: Ember.observer("searchText", function() {
     if (this.isValidTextLength()) {
       this.reloadResults();
+      this.get("cache").clear();
```
patrixr: you should probably move it before the `reloadResults`

patrixr: actually, best to put it *inside*  `reloadResults()`

patrixr: and actually, you can't use `this.get("cache")` here. Because it's declared in the controller, but not the mixin. The mixin cannot _assume_ that there is a cache

patrixr: so in the _controller_

You can do

```
reloadResults() {
   this.get("cache").clear();
   this._super();
}
```

app/controllers/orders/index.js
```diff
@@ -50,6 +56,11 @@ export default Ember.Controller.extend(SearchMixin, {
     };
   },
 
+  createCacheKey(key) {
+    const cacheKey = JSON.stringify(key);
+    return cacheKey;
+  },
```
patrixr:
```suggestion
  createCacheKey(data) {
    return JSON.stringify(data);
  },
```

app/controllers/orders/index.js
```diff
@@ -67,11 +79,15 @@ export default Ember.Controller.extend(SearchMixin, {
           this.getPaginationQuery(pageNo)
         )
       );
-
+      const cacheKey = this.createCacheKey(params);
+      if (cache.has(cacheKey)) {
+        return cache.get(cacheKey).value;
```
patrixr:
```suggestion
        return cache.get(cacheKey);
```

PR #674, opened 2020-03-20
GCW-3073-Message Box is displaying message with id and name as expected

shreyas098:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3073

### What does this PR do?

FEATURE: Add details about selected offer to message box 

Previously there was not any information about selected offer in the message box, so found the solution and added dependant information for it

### Screenshots

Screen Shot 2020-03-20 at 12 24 08 PM


app/components/goodcity/offers-search-overlay.js
```diff
@@ -25,6 +25,53 @@ export default Ember.Component.extend(SearchMixin, {
...
+          offerId: offer.get("firstObject.id")
+        }),
+        "Yes",
+        () => {
+          this.send("selectOffer", offer.get("firstObject"));
+        },
+        "No"
+      );
+    }
+  },
```
patrixr: When you end up doing a series of `if/else` in a row with big blocks, there's usually simpler/cleaner ways to write it.

For example

```javascript
  nameDisplayed(offer) {
    let data  = { offerId: offer.get('id') };
    let key   = 'search_offer.offer_select_warning_with_id';

    if (offer.get("createdBy")) {
      data['offerName'] = offer.get("createdBy.firstName") + offer.get("createdBy.lastName");
      key = 'search_offer.offer_select_warning_with_name_id';

      if (offer.get("company")) {
        data['offerCompany'] = offer.get("company.name");
        key = 'search_offer.offer_select_warning';
      }
    }

    this.get("messageBox").custom(
      this.get("i18n").t(key, data),
      "Yes",
      () => this.send("selectOffer", offer),
      "No"
    );
  }
```
Also

* `nameDisplay` doesn't tell us much on what it does. Maybe something like `confirmAssignOffer`
* There's no reason to pass an array of offers to a method which only applies to one offer. You can just pass the single offer to the method

app/components/goodcity/offers-search-overlay.js
```diff
@@ -50,14 +97,7 @@ export default Ember.Component.extend(SearchMixin, {
         .query("offer", params)
         .then(offers => {
           if (this.get("isMobileApp") && offers.get("length") == 1) {
```
patrixr: The idea that a pagination query method triggers a popup for a **specific** action INSIDE a **reusable** component is a big no-no. That logic belongs to the code that _triggers_ the overlay. (separation of concerns)

It needs to be fixed asap, because as of now this component is unusable anywhere else in the app.

I shouldn't have been like this to begin with. You can check with @abulsayyad123, he knows that code and he might know why this exists. Feel free to drop me a message if there's any questions :)

app/components/goodcity/offers-search-overlay.js
shreyas098: Thanks for the suggestion @patrixr . Implemented the above said part

app/components/goodcity/offers-search-overlay.js
shreyas098: Will discuss with @abulsayyad123 for the same

PR #680, opened 2020-03-30
GCW-3087 fixed location search at gain action

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3087

### What does this PR do?
This PR fixes issue with a location search at gain action form.

### DIFF
--- a/app/components/goodcity/locations-search-overlay.js
+++ b/app/components/goodcity/locations-search-overlay.js
@@ -41,6 +41,17 @@ export default SingletonComponent.extend(SearchMixin, {
       .slice(0, 10);
   }),
 
+  isLocationPresent(searchText) {
+    return function(location) {
+      return (
+        location
+          .get("name")
+          .toLowerCase()
+          .indexOf(searchText) > -1
+      );
+    };
+  },
+
   actions: {
     cancel() {
       this.send("selectLocation", null);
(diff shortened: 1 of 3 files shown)


app/components/goodcity/locations-search-overlay.js
```diff
@@ -51,11 +57,25 @@ export default SingletonComponent.extend(SearchMixin, {
     },
 
     loadMoreLocations(pageNo) {
-      const params = this.trimQuery(
-        _.merge({}, this.getSearchQuery(), this.getPaginationQuery(pageNo))
-      );
+      if (this.isValidTextLength()) {
```
abulsayyad123: We can return at start if text length is not valid 
```
 if (this.isValidTextLength()) {
    return ;
 }
```

app/components/goodcity/locations-search-overlay.js
```diff
@@ -51,11 +57,25 @@ export default SingletonComponent.extend(SearchMixin, {
...
+        if (this.get("presetLocations.length")) {
+          let searchText = this.getSearchQuery().searchText.toLowerCase();
+          return this.get("presetLocations").filter(function(location) {
```
abulsayyad123: We can move filter callback function to separate function.
```
  if (this.get("presetLocations.length")) {
      return this.get("presetLocations").filter(this.isLocationPresent);
  else {
      -----
  } 
```
For `isLocationPresent` method:
```
  isLocationPresent(location) {
     const searchText = this.getSearchQuery().searchText.toLowerCase();
     return (
              location
                .get("name")
                .toLowerCase()
                .indexOf(searchText) > -1
            );
  }
```
This becomes more readable.

app/components/goodcity/locations-search-overlay.js
```diff
@@ -51,11 +68,22 @@ export default SingletonComponent.extend(SearchMixin, {
...
-        _.merge({}, this.getSearchQuery(), this.getPaginationQuery(pageNo))
-      );
+      if (!this.isValidTextLength()) {
+        return;
+      }
+
+      if (this.get("presetLocations.length")) {
+        let searchText = this.getSearchQuery().searchText.toLowerCase();
+        return this.get("presetLocations").filter(
+          this.isLocationPresent(searchText)
```
abulsayyad123, line 78: No need to pass the argument `searchText` it automatically passes.

abulsayyad123, line 78: I think it would be better to access `searchText` in `isLocationPresent` method itself 🤔  instead of passing for every location.

PR #688, opened 2020-04-06
GCW-3103 - New item summary design

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3103

### What does this PR do?

In search results and other places where we show a summary version of an item we want to use a modified design that

 - Reflects the tabs visible on an item
       Links directly to each tab
 - Summarises available, designated & dispatched quantities

### Impacted Areas

1) Item search
2) Item details page (For Set type)


app/components/image-zoom.js
```diff
@@ -22,10 +23,10 @@ export default Ember.Component.extend({
 
   didInsertElement() {
     if (!this.get("isMobileApp")) {
-      var _this = this;
+      const parentId = `#${this.get("parentId")}`;
```
patrixr, line 26: We don't want to have any references to IDs in components. It makes them much harder to reuse.

I would recommend you use the jquery `.parent()` method to access the parent.

e.g

```
this.$().parent().lightGallery({
   ...
});
```

I don't really get why we need to access the parent to be honest, we could likely just create the gallery in the component's div.

app/templates/components/item-summary.hbs
```diff
@@ -0,0 +1,67 @@
+<section class="move-item-to-box">
```
patrixr, line 1: I believe this component already exist in the Actions UI built by @swatijadhav. Could you make sure there's no duplicate component ? You can probably use this component over there.

Also, a couple points

* `class="move-item-to-box"` makes little sense
* the name is very misleading, it's a `package` and not an `item`, and `"summary"` could mean anything. So I'd recomment `package-summary-block` for both the component name and the CSS class name

patrixr: PS: I know Stock uses the `item` wording instead of `package`. But it's wrong, and we need to start reversing that

abulsayyad123: Swati's UI is different than this. It is similar to box pallet UI. There was just HTML for that, I used that HTML (with modifications)  in this component.  
Surely, i'll change the CSS names.

app/components/image-zoom.js
abulsayyad123: Ahh.. it makes sense to use component default `div` (since components default tag is div). Will try it out and update and test it out. Thanks for pointing it out.

PR #705, opened 2020-04-14
GCW-3119 Add "Saleable" field to create / view / edit on packages

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3119

### What does this PR do?
Adds saleable field in creating and edit section of item

### Screenshots
image
image

Linked PR: https://github.com/crossroads/api.goodcity/pull/973

### DIFF
--- a/app/controllers/items/detail.js
+++ b/app/controllers/items/detail.js
@@ -9,7 +9,7 @@ import GradeMixin from "stock/mixins/grades_option";
 import MoveActions from "stock/mixins/move_actions";
 import DesignationActions from "stock/mixins/designation_actions";
 import StorageTypes from "stock/mixins/storage-type";
-import AsyncMixin from "stock/mixins/async";
+import AsyncMixin, { ERROR_STRATEGIES } from "stock/mixins/async";
 import ItemActions from "stock/mixins/item_actions";
 import _ from "lodash";
 import SearchMixin from "stock/mixins/search_resource";
@@ -440,6 +440,15 @@ export default GoodcityController.extend(
         const item = this.get("item");
         const saleable = item.get("saleable");
         item.set("saleable", !saleable);
+        this.runTask(async () => {
+          try {
+            await item.save();
+          } catch (e) {
+            item.rollbackAttributes();
+            throw e;
+          }
+        }, ERROR_STRATEGIES.MODAL);
+
         let loadingView = Ember.getOwner(this)
           .lookup("component:loading")
           .append();
--- a/app/controllers/items/detail.js
+++ b/app/controllers/items/detail.js
@@ -294,8 +294,20 @@ export default GoodcityController.extend(
       }
     },
 
-    myaction: Ember.observer("item.saleable", function() {
-      console.log("I will fire the update query");
+    onSaleableChange: Ember.observer("item.saleable", function() {
+      const item = this.get("item");
+      const previousValue = this.get("previousSaleableValue");
+      if (previousValue == null) {
+        this.set("previousSaleableValue", this.get("item.saleable"));
+        return;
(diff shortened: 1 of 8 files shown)


app/controllers/items/detail.js
```diff
@@ -294,6 +294,22 @@ export default GoodcityController.extend(
       }
     },
 
+    onSaleableChange: Ember.observer("item.saleable", function() {
```
patrixr, line 297: A few things:

* An observer for a checkbox feels a bit too agressive. I think having a "saveItem" action linked to the checkbox's `change` event is more appropriate.
* Having a `previousSaleableValue` value doesn't seem used in any meaningful way here. If there is an error you can do `item. rollbackAttributes()` to restore it as it was (I assume that's what you were going for)

Also, we have an "async" mixin you can use to avoid having to take care of the loading spinner.

e.g

```javascript
{
   saveItem() {
      this.runTask(async () => {
           try {
               await item.save();
           } catch (e) {
               item.rollbackAttributes();
               throw e;
           }
      }, this.ERROR_STRATEGIES.MODAL);
  }
}

// hbs 
{{input type="checkbox" checked=isSaleable change=(action "saveItem") }} // might be wrong about the syntax
```

app/templates/items/detail/tabs/_item_detail.hbs
```diff
@@ -78,6 +78,16 @@
           </div>
         </div>
       </div>
+      <div class="row detail-section">
+        <div class="columns small-12">
+          <div class="row">
+            <div class="columns small-3">Saleable </div>
+            <div class="columns small-9 value description">
+              {{input class="checkbox-item-details" type="checkbox" id="saleable" checked=item.saleable }}
```
patrixr, line 86: It's error-prone to have IDs on elements, especially in a reusable partial. I already see the same ID being used twice in this PR. As a best-practice we want to avoid them. And they're rarely very useful anyways :)

app/controllers/items/detail.js
bharat619: thanks @patrixr ,
`runTask` was awesome sleek. I have updated it, as well as created action instead of observers

app/templates/items/detail/tabs/_item_detail.hbs
bharat619: sure. Will remove the ids. Thanks @patrixr

app/controllers/items/detail.js
```diff
@@ -436,6 +436,27 @@ export default GoodcityController.extend(
...
+            throw e;
+          }
+        }, ERROR_STRATEGIES.MODAL);
+
+        let loadingView = Ember.getOwner(this)
+          .lookup("component:loading")
+          .append();
+        item.save().finally(() => {
+          loadingView.destroy();
+        });
```
abulsayyad123, line 457: Remove this entire line.It's already done above. `runTask` takes care of spinner.

bharat619: done 👍 
i had kept it by mistake. have removed it

app/templates/items/new.hbs
```diff
@@ -161,6 +161,18 @@
...
     </div>
+    <div class="row">
+      <div class="small-3 columns item-label">
+        Saleable
+      </div>
+      <div class="small-3 columns publish">
+        {{input type="checkbox"
+        checked=isSaleable}}
+      </div>
+      <div class="small-9 columns">
```
abulsayyad123, line 172: Minor thing.. you can remove this div. Because the next div is having `row` class it is anyhow going to take new line. Or next thing you can do is use  `offsett` class in second div with class `publish`.

PR #708, opened 2020-04-16
GCW-3120 Message Center

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3120

### What does this PR do?
Messages Center to display a list of all unread messages. 

### Screenshots
Screenshot 2020-04-16 at 3 31 21 PM

### Linked PR's:
https://github.com/crossroads/api.goodcity/pull/977

### DIFF
--- a/app/services/messages.js
+++ b/app/services/messages.js
@@ -1,30 +1,90 @@
 import Ember from "ember";
 const { getOwner } = Ember;
+import AjaxPromise from "stock/utils/ajax-promise";
 
 export default Ember.Service.extend({
   logger: Ember.inject.service(),
   session: Ember.inject.service(),
+  store: Ember.inject.service(),
 
-  markRead: function (message) {
+  unreadMessageCount: 0,
+
+  fetchUnreadMessageCount() {
+    return this._queryMessages("unread", 1, 1)
+      .then(data => {
+        const count = data.meta && data.meta.total_count;
+        this.set("unreadMessageCount", count || 0);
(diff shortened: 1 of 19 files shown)


app/services/messages.js
```diff
@@ -1,30 +1,90 @@
...
         .then(data => {
           delete data.message.id;
           message.setProperties(data.message);
+          this._decrementCount();
         })
         .catch(error => this.get("logger").error(error));
     }
   },
 
+  markAllRead() {
```
patrixr, line 45: @swatijadhav I think this endpoint might also mark all the Admin notifications as read 🤔 , which we don't want.

But I believe this endpoint takes in a `"context"` param, which you can set to `"order"`

patrixr: I guess we need to check that Admin does the same

app/templates/app_menu_list.hbs
```diff
@@ -51,4 +51,4 @@
     </ul>
   </div>
 </div>
-{{partial "sticky_footer"}}
+{{sticky-footer}}
```
patrixr, line 54: 👍  good move, I think a lot of the partials are more appropriate as components around the app. 
I think we should only use partials to divide templates which are too big, not so much for making reusable pieces

app/services/messages.js
swatijadhav: Thanks for pointing out this one. Changes done!

PR #723, opened 2020-04-24
GCW-3127 Place overridden value on Stock item screen

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3127

### What does this PR do?
Creates new field for value based on the grade of the package.

### Impacted Areas
Stock > Create new Item

### Screenshots
image
image

### DIFF
--- a/app/controllers/items/detail.js
+++ b/app/controllers/items/detail.js
@@ -169,15 +169,14 @@ export default GoodcityController.extend(
       "prevValueHkDollar",
       function() {
         const item = this.get("item");
-        const hasValueHkDollarChanged = Object.keys(
-          item.changedAttributes()
-        ).includes("valueHkDollar");
-        return (
-          hasValueHkDollarChanged &&
-          !!item.get("valueHkDollar") &&
-          parseFloat(this.get("prevValueHkDollar")) !==
-            parseFloat(item.get("valueHkDollar"))
-        );
+        const valueHkDollar = parseFloat(item.get("valueHkDollar"));
+        const prevValueHkDollar = parseFloat(this.get("prevValueHkDollar"));
+        const defaultValue = parseFloat(this.get("defaultValueHkDollar"));
+        if (!prevValueHkDollar) {
+          return Math.abs(valueHkDollar - defaultValue);
+        } else {
+          return Math.abs(valueHkDollar - prevValueHkDollar);
+        }
       }
     ),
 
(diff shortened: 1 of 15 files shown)


app/controllers/items/detail.js
```diff
@@ -154,6 +154,15 @@ export default GoodcityController.extend(
       }
     }),
 
+    canApplyDefaultValuation: Ember.computed("model.valueHkDollar", function() {
+      const valueHkDollar = +this.get("model.valueHkDollar");
+      let defaultValue = +this.get("defaultValueHkDollar");
+      defaultValue = !!defaultValue
+        ? defaultValue
+        : this.set("defaultValueHkDollar", valueHkDollar);
```
patrixr, line 162: doing a `set` in a computed property is always a code smell. Better keep those safe and readonly, state should be modified at initialisation or through user actions

app/controllers/items/new.js
```diff
@@ -114,6 +115,17 @@ export default GoodcityController.extend(
       }
     }),
 
+    canApplyDefaultValuation: Ember.computed("valueHkDollar", function() {
+      const defaultValue = this.get("defaultValueHkDollar");
+      const valueHkDollar = this.get("valueHkDollar");
+
+      if (!defaultValue) {
+        this.send("getItemValuation");
```
patrixr, line 123: Same comment as above, this modifies the state during a read.

Also note that `getXXX` is a standard naming format for getters, most people would assume it's a safe read.
Better to name it something like `updateValuation` which explicitly states that it will modify something

app/controllers/items/detail.js
bharat619: Thanks @patrixr , This was a ToDo and I had missed it. Thanks for pointing it out

app/controllers/items/new.js
bharat619: thanks. its done

app/controllers/items/detail.js
```diff
@@ -391,6 +418,27 @@ export default GoodcityController.extend(
...
+        this.send("saveItem", item);
+      },
+
+      /**
+       * Updates the valueHkDollar
+       * Updates the previous saved value
+       */
+      async updateItemValuation() {
+        const item = this.get("item");
+        this.send("saveItem", item);
```
abulsayyad123, line 438: You should await here 🤔 ?

bharat619: not needed. Removed the async thing

PR #724, opened 2020-04-27
Gcw 3111 add duplicate item checkbox

shreyas098:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3111

### What does this PR do?

FEATURE:To duplicate creation of item

### DIFF
--- a/app/components/dropdown-search.js
+++ b/app/components/dropdown-search.js
@@ -23,23 +23,36 @@ export default Ember.Component.extend({
   displayLabel: Ember.computed("addAble", function() {
     return this.get("addAble") ? this.get("addItem") : "";
   }),
+  didRender() {
+    if (this.isDuplicateField) {
+      console.log(this.content);
+      this.send("addNewField", this.content, null);
+    }
+  },
 
   selectedOptionDisplay: Ember.computed("dropDownValues", function() {
     let selectedValues = {
       ...this.get("dropDownValues")
     };
+    console.log(selectedValues);
     let dataObj = {};
     Object.keys(selectedValues).map((data, index) => {
       if (data != "country_id") {
         if (this.get("fixedDropdownArrId").indexOf(data) >= 0) {
+          console.log(data);
           let field = `${data.substring(0, data.length - 3)}`;
+          console.log(field);
           let recordData = this.get("store")
             .peekRecord("lookup", selectedValues[data])
             .get("labelEn");
+          console.log(selectedValues[data]);
+          console.log(recordData);
+          console.log(this.isDuplicateField);
           dataObj[field] = {
             id: selectedValues[data],
             tag: recordData
           };
+          console.log(dataObj);
         } else {
           dataObj[data] = {
             id: index + 1,
@@ -98,6 +111,7 @@ export default Ember.Component.extend({
       packageDetails[name].push(newTag);
       this.set("packageDetails", packageDetails);
       this.send("setSelected", name, newTag);
+      this.set("isDuplicateField", false);
     },
 
     async setSelected(fieldName, value) {
--- a/app/components/dropdown-search.js
+++ b/app/components/dropdown-search.js
@@ -25,6 +25,7 @@ export default Ember.Component.extend({
   }),
 
   didUpdate() {
+    this._super(...arguments);
     if (this.isDuplicateField) {
       let dropDownValues = {
         ...this.get("dropDownValues")
@@ -112,7 +113,6 @@ export default Ember.Component.extend({
       packageDetails[name].push(newTag);
       this.set("packageDetails", packageDetails);
       this.send("setSelected", name, newTag);
-      this.set("isDuplicateField", false);
     },
 
     async setSelected(fieldName, value) {
--- a/app/components/dropdown-search.js
+++ b/app/components/dropdown-search.js
@@ -24,16 +24,6 @@ export default Ember.Component.extend({
     return this.get("addAble") ? this.get("addItem") : "";
   }),
 
-  didUpdate() {
-    this._super(...arguments);
-    if (this.isDuplicateField) {
-      let dropDownValues = {
-        ...this.get("dropDownValues")
-      };
-      this.set("dropDownValues", dropDownValues);
-    }
-  },
-
   selectedOptionDisplay: Ember.computed("dropDownValues", function() {
     let selectedValues = {
       ...this.get("dropDownValues")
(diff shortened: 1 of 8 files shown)


app/components/dropdown-search.js
```diff
@@ -24,6 +24,16 @@ export default Ember.Component.extend({
     return this.get("addAble") ? this.get("addItem") : "";
   }),
 
+  didUpdate() {
+    this._super(...arguments);
+    if (this.isDuplicateField) {
```
patrixr, line 29: we should _never_ refer to specific features in generic/reusable components !
Plus, this piece of code doesn't seem to really be doing anything

app/controllers/items/new.js
```diff
@@ -93,6 +104,10 @@ export default GoodcityController.extend(
       }
     ),
 
+    showDuplicateCheckbox: Ember.computed("storageType", function() {
+      return ["Package"].indexOf(this.get("storageType")) > -1;
+    }),
```
patrixr, line 109:
```suggestion
    showDuplicateCheckbox: Ember.computed.equal("storageType", "Package"),
```

app/services/settings.js
```diff
@@ -83,6 +84,12 @@ export default Ember.Service.extend({
     return !this.readBoolean("stock.allow_box_pallet_item_addition");
   }),
 
+  /**
+   * @property {Computed<string>} paramsNotCopied lists params not to be copied during duplicate items
+   */
+  paramsNotCopied: Ember.computed(function() {
+    return this.readString("stock.params_not_to_be_copied");
```
patrixr, line 91: You can probably add a `readArray` or `readList` method, that way it's reusable and you can remove the splitting code from the controller

app/controllers/items/new.js
```diff
@@ -177,6 +192,32 @@ export default GoodcityController.extend(
       return detailAttributes;
     },
 
+    paramsToBeCleared() {
```
patrixr, line 195: The name `clearParams` would make more sense since the method actually does the clear. The current name would indicate that it returns the list of params to be cleared.

app/controllers/items/new.js
```diff
@@ -463,13 +504,29 @@ export default GoodcityController.extend(
...
-            locationId: "",
-            inventoryNumber: "",
-            offersLists: []
-          });
-          this.replaceRoute("items.detail", data.item.id);
+          if (this.get("isDuplicate") && !this.get("isBoxOrPallet")) {
+            this.replaceRoute("items.new");
+            this.paramsToBeCleared();
+            this.set("quantity", 1);
+            this.send("autoGenerateInventoryNumber");
```
patrixr, line 511: I'm guessing this method returns a promise (please double check), if so you likely want to await that method

app/controllers/items/new.js
```diff
@@ -463,13 +504,29 @@ export default GoodcityController.extend(
...
           }
           this.updateStoreAndSaveImage(data);
-          this.clearSubformAttributes();
-          this.setProperties({
-            locationId: "",
-            inventoryNumber: "",
-            offersLists: []
-          });
-          this.replaceRoute("items.detail", data.item.id);
+          if (this.get("isDuplicate") && !this.get("isBoxOrPallet")) {
```
patrixr, line 507: Not sure about the `isDuplicate` name, doesn't `shouldDuplicate` make more sense ? As in, when we check the box, we indicate that we _want_ to create a duplicate. Not that this item _is_ a duplicate of anything 🤔

shreyas098: sure will change that

app/controllers/items/new.js
shreyas098: sure

app/controllers/items/new.js
shreyas098: sure will do that

app/components/dropdown-search.js
shreyas098: So this is basically done to reset the dropdown values to null when the page is reloaded

shreyas098: And in order to protect the reusability i have passed params IsDuplicate for it

patrixr:
```
let dropDownValues = {
        ...this.get("dropDownValues")
      };
```

This wouldn't set anything to null 🤔 

Bottom line is, we can't have any reference to the duplicate feature in this file. You can probably trigger that from the controller somehow

app/services/settings.js
shreyas098: Thanks ..will do that

app/controllers/items/new.js
shreyas098: yes added await for the required methods

app/templates/items/new.hbs
```diff
@@ -240,6 +232,24 @@
...
+      <div class="row">
+        {{#if showPublishItemCheckBox}}
+          <div class="small-3 columns item-label">
+            <label for="publishItem">{{t "items.new.publish"}}</label>
+          </div>
+          <div class="small-3 columns publish">
+            {{input type="checkbox" id="publishItem" checked=isAllowedToPublish}}
+          </div>
+        {{/if}}
+        {{#if showDuplicateCheckbox}}
```
abulsayyad123, line 244: indentation.

app/controllers/items/new.js
```diff
@@ -177,6 +189,30 @@ export default GoodcityController.extend(
...
     },
 
+    clearParams() {
+      if (this.get("displayFields")) {
+        let attr = this.get("paramsNotCopied");
+        let fieldAttributes = this.get("displayFields").map(
+          value => value.name
+        );
+        attr.forEach(value => {
+          if (fieldAttributes.indexOf(value) > -1) {
```
abulsayyad123, line 199: Just a suggestion.. move this iteration logic to other function to make it more readable like 
```suggestion
        attr.forEach(this.clearAttributes)
```
```
        clearAttributes(value) {
          //write if-else or switch statement here.
        }

File: app/controllers/items/new.js [hunk 9]
```diff
@@ -177,6 +189,30 @@ export default GoodcityController.extend(
       return detailAttributes;
     },
 
+    clearParams() {
+      if (this.get("displayFields")) {
+        let attr = this.get("paramsNotCopied");
+        let fieldAttributes = this.get("displayFields").map(
+          value => value.name
+        );
+        attr.forEach(value => {
```
abulsayyad123, line 198: Just a suggestion.. move this iteration logic to other function to make it more readable like

```suggestion
        attr.forEach(this.clearAttribute)
```

app/templates/items/new.hbs
shreyas098: sure will fix it

PR #738, opened 2020-05-12
GCW-3128- Add Package Type selection overlay.

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3128

### What does this PR do?
This PR has 
- It applies highlighted package_type selection overlay on pages where it was not applied. 
- Removed `search_code` pages for package_type selection 
- Refactored Asynchronous spinner code.
- Fixed failing specs.


app/components/add-request.js
```diff
@@ -37,20 +38,24 @@ export default Ember.Component.extend({
...
+
+      if (pkgType) {
+        this.runTask(
+          this.get("goodcityRequest").updateGcRequest(reqId, {
+            package_type_id: pkgType.get("id"),
+            quantity: 1,
+            order_id: this.get("order.id")
+          })
+        );
+      }
```
patrixr, line 58: I feel the name `goodcityRequest` is a bit misleading. Doesn't indicate that it's a service.

I'd also recommend renaming `getPackageType` to `userPickPackageType` to be consistent with the other methods that trigger UI overlays. Also more explicit

app/controllers/items/detail.js
```diff
@@ -337,6 +336,84 @@ export default GoodcityController.extend(
...
+          packageName
+        }),
+        translation.t("not_now"),
+        null,
+        translation.t("continue"),
+        () => {
+          this.deleteAndAssignNew(pkgType);
+        }
+      );
+    },
```
patrixr, line 371: This method is really verbose and not very useful. You could perhaps add a helper for that (maybe in the async service??).

e.g

```javascript
this.withConfirmation('"items.new.subform.delete_subform_waring").runTask(() => {
  return this.deleteAndAssignNew(pkgType);
});
```

app/controllers/items/detail.js
```diff
@@ -402,6 +479,15 @@ export default GoodcityController.extend(
         );
       },
 
+      async getPackageType() {
```
patrixr, line 482: `getPackageType` as a name also doesn't make a lot of sense here.

app/components/add-request.js
abulsayyad123: Done!

PR #741, opened 2020-05-15
GCW-3145 Fix expiry_date rollbar error

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3145

`expiry_date` was being passed as attribute for under `detail_attributes` and was causing this error.

This fix will only allow the params that are part of display fields to be passed under `detail_attributes`


bharat619:
Closing this PR as issue is fixed in https://github.com/crossroads/api.goodcity/pull/1008

PR #743, opened 2020-05-19
GCW-3101 Package sets

patrixr:
Support for Package Sets

Changes:
* Removal of `setItem` references
* Addition of `PackageSet` model
* Packages are now grouped via that packageSet
* Addition of an "Action menu" on item details page
* New set actions: "Make set", "Unlink" and "Add to set"
* Refactoring of package and packageType overlays
* Fix for Tim's comment on GCW-3104


app/components/reactive-view.js
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Re-renders it's content on model change
+ *
+ * @module Components/reactive-view
+ * @augments ember/Component
+ * @property {any} model a value to listen to
+ *
+ * @example
+ *
+ * {{#reactive-view model=something}}
```
abulsayyad123, line 12: 👍

app/helpers/chain-actions.js
```diff
@@ -0,0 +1,25 @@
...
+ *  )
+ * }}>
+ *  Submit
+ * </button>
+ *
+ */
+export default Ember.Helper.helper(function(actions) {
+  return async function() {
+    for (let action of actions) {
+      await action();
```
abulsayyad123, line 22: Exceptions will be handled at `action` level ?

patrixr: Yeah, although you do have a point that it would carry on the chain regardless of errors. In which case it might be worth renaming to "multiple-actions" not to infer that it handles chain errors

PR #751, opened 2020-05-26
GCW-3133 Manage Users Screens

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3133

### What does this PR do?

Screenshot 2020-05-26 at 7 12 49 PM

Screenshot 2020-05-29 at 2 31 39 PM

Screenshot 2020-05-26 at 7 23 21 PM


app/components/check-box.js
```diff
@@ -0,0 +1,28 @@
+import Ember from "ember";
+
+export default Ember.Component.extend({
```
patrixr, line 3: do we not have a checkbox component already ? :(

app/controllers/add_user.js
```diff
@@ -0,0 +1,281 @@
...
+      organisations_users_ids: [this.get("organisationId")]
+    };
+  },
+
+  saveImage() {
+    let image = this.get("newUploadedImage");
+    if (image) {
+      return image.save();
+    } else {
+      return Promise.resolve();
```
patrixr, line 97: tip: if you make it an async method, there's no need to manually do a `Promise.resolve` :)

app/controllers/add_user.js
```diff
@@ -0,0 +1,281 @@
...
+          .catch(xhr => {
+            if (xhr.status === 422) {
+              this.get("messageBox").alert(xhr.responseJSON.errors[0].message);
+            } else {
+              throw xhr;
+            }
+          })
+          .finally(() => loadingView.destroy());
+      });
+    },
```
patrixr, line 159: Using the `async` mixin we have, and an async method you could significantly reduce the size of a lot of those methods, making them more readable. 

As an example:

```javascript
  saveUser() {
    let newUser = this
      .get("store")
      .createRecord("user", this.getRequestParams());

    newUser.set("userRoleIds", this.getWithDefault("selectedRoleIds", []));

    return this.runTask(async () => {
      newUser.set("image", await this.saveImage());

      const { id } = await newUser.save();

      this.clearFormData();
      this.transitionToRoute("user_details", id);

    }, this.ERROR_STRATEGIES.MODAL);
  }
```

we basically don't want manually creating the spinner anywhere, or the error modal.

Note: doing `let _this = this` is unnecessary, lambda methods use the current scope

app/controllers/add_user.js
```diff
@@ -0,0 +1,281 @@
...
+        data.result.public_id +
+        "." +
+        data.result.format;
+      var newUploadedImage = this.get("store").createRecord("image", {
+        cloudinaryId: identifier,
+        favourite: true
+      });
+      this.set("newUploadedImage", newUploadedImage);
+      this.set("userImageKeys", identifier);
+    },
```
patrixr, line 222: All the upload methods must certainly been written over and over again, can we maybe centralise them in a mixin ?

app/controllers/search_organisation.js
```diff
@@ -1,6 +1,8 @@
 import Ember from "ember";
 
 export default Ember.Controller.extend({
+  queryParams: ["backToNewUser"],
+  backToNewUser: false,
```
patrixr, line 5: Can we use normal browser back behaviour, we've been trying to step away from these kinds of hacks.
If there's a reason we can't, then at least let's make it usable for any page with a `redirectTo` param which can support any page

app/controllers/manage_users.js
```diff
@@ -0,0 +1,49 @@
...
+        roles: [
+          "Charity",
+          "Supervisor",
+          "Reviewer",
+          "Order fulfilment",
+          "Order administrator"
+        ]
+      };
+
+      if (this.get("searchText").trim().length >= 3) {
```
patrixr, line 44: The length requirement shouldn't be present in a `loadMore` method, as that is passed on to the component which expects an array returned.

Better move that to `onSearchTextChange`, and not trigger the search at all in the case where the we're under 3 characters

app/components/check-box.js
```diff
@@ -0,0 +1,28 @@
...
+  attributeBindings: ["type", "value"],
+  tagName: "input",
+  type: "checkbox",
+  checked: false,
+  selection: [],
+
+  didInsertElement() {
+    var isCurrentRole =
```
patrixr, line 11: Components should be generic and reusable, so better not use the term `role` at all. Maybe substitute `isCurrentRole` by `isChecked` ?

app/components/check-box.js
swatijadhav: no, it was not there

app/components/async-validatable-form.js
```diff
@@ -0,0 +1,19 @@
+import ValidatableForm from "ember-cli-html5-validation/components/validatable-form";
+import AsyncMixin, { ERROR_STRATEGIES } from "stock/mixins/async";
+
+export default ValidatableForm.extend(AsyncMixin, {
```
patrixr, line 4: I think @abulsayyad123  has done some work regarding forms and validation, could you guys sync up to make sure we're doing it in a consistent way.

app/controllers/add_user.js
```diff
@@ -0,0 +1,158 @@
+import config from "stock/config/environment";
+import AsyncMixin, { ERROR_STRATEGIES } from "stock/mixins/async";
+import ImageUploadMixin from "stock/mixins/image_upload";
```
patrixr, line 3: <3

app/components/async-validatable-form.js
abulsayyad123: It is not completed yet.. some minor work is remaining. Once it is done we can later start using it. This looks fine we can with this for now.

app/templates/user_details.hbs
```diff
@@ -0,0 +1,144 @@
...
+        <div class="columns small-8 text">
+          {{time-ago timeValue=model.lastDisconnected smallPrefix=false}}
+        </div>
+      </div>
+    </div>
+
+    <div class="data-rows">
+      <div class="row">
+        <div class="columns small-4 text-prefix">
+          Disabled:
```
abulsayyad123, line 131: translation missing

PR #759, opened 2020-06-01
GCW-3153-Add constraints restriction to packages

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3153

### What does this PR do?

This PR adds constraints, saleable and comment field in packages creation and details page.

### DIFF
--- a/app/constants/saleable-options.js
+++ b/app/constants/saleable-options.js
@@ -1,0 +1,5 @@
+export const SALEABLE_OPTIONS = [
+  { id: "Not set", name: "Not set", value: null },
+  { id: "Saleable", name: "Saleable", value: "true" },
+  { id: "Not selling", name: "Not selling", value: "false" }
+];
--- a/app/constants/saleable-options.js
+++ b/app/constants/saleable-options.js
@@ -1,5 +1,9 @@
 export const SALEABLE_OPTIONS = [
-  { id: "Not set", name: "Not set", value: null },
-  { id: "Saleable", name: "Saleable", value: "true" },
-  { id: "Not selling", name: "Not selling", value: "false" }
+  { id: "Not Set", translation_key: "items.new.not_set", value: null },
+  { id: "Saleable", translation_key: "items.new.saleable", value: "true" },
+  {
+    id: "Not selling",
+    translation_key: "items.new.not_selling",
+    value: "false"
+  }
 ];
--- a/app/controllers/items/new.js
+++ b/app/controllers/items/new.js
@@ -17,7 +17,19 @@ export default GoodcityController.extend(
   GradeMixin,
   AsyncMixin,
   {
+<<<<<<< HEAD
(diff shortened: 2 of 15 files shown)


app/constants/saleable-options.js
```diff
@@ -0,0 +1,5 @@
+export const SALEABLE_OPTIONS = [
+  { id: "Not set", name: "Not set", value: null },
+  { id: "Saleable", name: "Saleable", value: "true" },
+  { id: "Not selling", name: "Not selling", value: "false" }
+];
```
patrixr, line 5: Can you add those to the translation file ?

app/controllers/items/new.js
```diff
@@ -76,6 +76,14 @@ export default GoodcityController.extend(
     }),
     offersLists: [],
 
+    saleableId: Ember.computed("saleableOptions", function() {
+      return this.get("saleableOptions").get("firstObject");
+    }),
+
+    restrictionId: Ember.computed("restrictionOptions", function() {
+      return this.get("restrictionOptions").get("firstObject");
+    }),
```
patrixr, line 85: Since you're modifying those, it's better not to use a computed property, you can simply initialise them

app/constants/saleable-options.js
```diff
@@ -0,0 +1,9 @@
+export const SALEABLE_OPTIONS = [
+  { id: "Not Set", translation_key: "items.new.not_set", value: null },
+  { id: "Saleable", translation_key: "items.new.saleable", value: "true" },
```
patrixr, line 3: how come the booleans are strings ?

abulsayyad123: if i am going with boolean the null field is converted to false even if i want to set null.

app/mixins/grades_option.js
```diff
@@ -1,11 +1,25 @@
...
       name: key,
       id: key
     }));
+  }),
+
+  saleableOptions: Ember.computed(function() {
+    return _.map(SALEABLE_OPTIONS, obj => {
+      return { ...obj, name: this.get("i18n").t(obj.translation_key).string };
+    });
+  }),
```
patrixr, line 20: this computed property should be dependant on the current language, otherwise it won't change when we change the language of the app

abulsayyad123: On changing the current language whole app get refreshed so it will take the change.

app/controllers/items/new.js
```diff
@@ -349,6 +349,8 @@ export default GoodcityController.extend(
         expiry_date: this.get("expiry_date"),
         value_hk_dollar: this.get("valueHkDollar"),
         offer_ids: this.get("offersLists").getEach("id"),
+        restriction_id: this.get("restrictionId").id,
+        saleable: this.get("saleableId").value,
```
patrixr, line 353: You can remove those two lines

PR #767, opened 2020-06-08
GCW-3170 Staff chat to orders

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3170

### What does this PR do?
Staff chat to orders

Screenshot 2020-06-04 at 10 08 12 AM
Screenshot 2020-06-04 at 11 36 26 AM


app/controllers/application.js
```diff
@@ -24,8 +25,18 @@ export default Ember.Controller.extend({
       // jshint ignore:line
       this.redirectToItem();
     }
+
+    this.get("subscription").on("change:message", this, this.onNewNotification);
   }),
 
+  onNewNotification(notification) {
```
abulsayyad123, line 32: Can we use `onNewNotification({id}=notification.reocrd) {` just to fetch id 🤔

app/controllers/orders/conversation.js
```diff
@@ -68,8 +69,9 @@ export default detail.extend({
     var message = this.store.createRecord("message", values);
     message
       .save()
-      .then(() => {
+      .then(data => {
         this.set("body", "");
+        this.get("messages").pushObject(data._internalModel);
```
abulsayyad123, line 74: Why there is need to get `_internalModel`? I find fetching data from Ember data though `_internalModel` very fishy 😄 .

app/controllers/orders/index.js
```diff
@@ -23,6 +23,12 @@ export default Ember.Controller.extend(SearchMixin, {
     this.cache = new Cache();
   },
 
+  searchProps: {
```
abulsayyad123, line 26: Are we fetching `messages` in `index` page as well ?

app/routes/orders/staff_conversation.js
```diff
@@ -0,0 +1,19 @@
+import conversation from "./conversation";
+import Ember from "ember";
+
+export default conversation.extend({
+  templateName: "orders/conversation",
+
+  model(params) {
+    return Ember.RSVP.hash({
+      designation:
+        this.store.peekRecord("designation", params.order_id, {
```
abulsayyad123, line 10: I think `loadIfAbsent` would be better to use here.

app/controllers/orders/index.js
swatijadhav: Yes, we will only fetch messages (not message associations) to display the count of unread messages on each order.

Previously, gave a try to fetch count of unread messages from the API serializer response, but again we need to handle live update of that count on new message and also when messages are marked read. But that will add unnecessary observers, hence fetching messages records directly and live-update will be auto handled.

app/controllers/application.js
```diff
@@ -24,8 +25,18 @@ export default Ember.Controller.extend({
...
+    const msg = this.get("store").peekRecord("message", notification.record.id);
+
+    if (msg.get("isUnread")) {
+      this.get("messagesUtil")._incrementCount();
```
patrixr, line 36: `_incrementCount` shouldn't be called from outside the service. It's meant as a private method (thus the leading `_`)

app/controllers/orders/conversation.js
patrixr: agreed

swatijadhav: Since pushing `data` directly gives error:
```js
Uncaught TypeError: internalModel.getRecord is not a function
    at Class.objectAtContent (-private.js:9360)
    at Class.objectAt (array_proxy.js:233)
    at objectAt (array.js:42)
    at addObserverForContentKey (array.js:399)
    at EachProxy.arrayDidChange (array.js:353)
    at arrayContentDidChange (array.js:118)
    at Class._emberMetal.Mixin.create._Mixin$create.arrayContentDidChange (array.js:286)
    at Class.arrangedContentArrayDidChange (array_proxy.js:345)
    at Object.applyStr (ember-utils.js:529)
    at sendEvent (ember-metal.js:425)
```
Hence using `(data._internalModel);` which suffices our purpose here.

PR #770, opened 2020-06-09
GCW-3167 User mentions

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3167

### What does this PR do?
Adds an UI ability to mention related users

### Impacted Areas
The chat and notification sections

### Screenshots
image
image


app/controllers/orders/conversation.js
```diff
@@ -24,7 +24,6 @@ export default detail.extend({
   sortedMessages: Ember.computed.sort("model.messages", "sortProperties"),
 
   groupedMessages: Ember.computed("sortedMessages", function() {
-    this.autoScroll();
```
patrixr, line 27: 👍

patrixr:
@bharat619 Looks good 👍  Although I don't see where tributejs is being used, could you explain how it's being used ?

bharat619:
@patrixr , 
The tributejs is used in the message-textbox addon https://github.com/crossroads/message-textbox
This is added to the goodcity-lib PR https://github.com/crossroads/goodcity-lib/pull/47
So browse and stock will be reading from that

More details:
I have added `ember-browserify` dev dependency to directly include the npm packages.
On using npm package with browse add on will simply not consume an npm dependency for the host app. This is a limitation of ember-cli
Have follwed the workaround mentioned in this step
https://www.npmjs.com/package/ember-browserify#using-ember-browserify-in-addons

PR #775, opened 2020-06-11
GCW- 3173-Item creation datepicker UX

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3173

### What does this PR do?

This PR adds feature to select years from dropdown in datepicker instead of navigating through different months.

### DIFF
--- a/app/components/calendar-input.js
+++ b/app/components/calendar-input.js
@@ -1,7 +1,10 @@
 import Ember from "ember";
 import _ from "lodash";
 
-const PICKADATE_CONFIG = {
+const MIN_DATE = moment()
+  .subtract(2, "years")
+  .toDate();
+const DEFAULT_PICKADATE_CONFIG = {
   selectMonths: true,
   selectYears: true,
   format: "ddd mmm d",
@@ -70,11 +73,13 @@ export default Ember.TextField.extend({
 
   didReceiveAttrs() {
     this._super(...arguments);
-    const enablePreviousDate = this.get("enablePreviousDate");
-    if (enablePreviousDate) {
-      PICKADATE_CONFIG.min = moment()
-        .subtract(2, "years")
-        .toDate();
+    this.set("datePickerConfig", DEFAULT_PICKADATE_CONFIG);
+    const enablePastDate = this.get("enablePastDate");
+    if (enablePastDate) {
+      this.set("datePickerConfig", {
+        ...DEFAULT_PICKADATE_CONFIG,
+        min: MIN_DATE
+      });
     }
   },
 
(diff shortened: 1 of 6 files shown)


app/components/calendar-input.js
```diff
@@ -66,6 +68,16 @@ export default Ember.TextField.extend({
     }
   },
 
+  didReceiveAttrs() {
+    this._super(...arguments);
+    const enablePreviousDate = this.get("enablePreviousDate");
+    if (enablePreviousDate) {
+      PICKADATE_CONFIG.min = moment()
```
patrixr, line 75: an all caps variable is a convention for a constant. So modifying it is a big no no.

Especially given that this object is _shared_ across all `calendar-input` component instanciation. That means that if only one modifies it, all the others in the future will have that modifictation.

I would suggest having a constant for the default config, and then extending it

e.g

```javascript

const MIN_DATE = moment().subtract(2, "years").toDate();
const DEFAULT_PICKADATE_CONFIG = {
   // something
};

// This is just a helper to avoid doing the ternaries that you're doing in date-picker.js. Thought it was nicer
const optional = (val) => ({
   if: (cond) => cond ? val : undefined
});

// ...

{

  didReceiveAttrs() {
    this._super(...arguments);
    const enablePreviousDate = this.get("enablePreviousDate");
    this.set('datePickerConfig', {
       ...DEFAULT_PICKADATE_CONFIG,
       min: optional(MIN_DATE).if(enablePreviousDate)
    });

    // and then you use `datePickerConfig` inside the component
  }
}
```

app/components/date-picker.js
```diff
@@ -36,13 +36,21 @@ export default Ember.TextField.extend({
...
+        selectYears: !!enablePreviousDate,
         format: "ddd mmm d",
         monthsFull: moment.months(),
         monthsShort: moment.monthsShort(),
         weekdaysShort: moment.weekdaysShort(),
+        min: enablePreviousDate
+          ? moment()
+              .subtract(2, "years")
+              .toDate()
+          : undefined,
```
patrixr, line 53: Using the helper I wrote above you could do

```javascript
min: optional(/* the moment */).if(enablePreviousDate)
```

It's not a change request, I just think there are ways to make things more pleasant to read :)

app/components/date-picker.js
```diff
@@ -36,13 +36,21 @@ export default Ember.TextField.extend({
 
   didInsertElement() {
     const cmp = this;
+    const enablePreviousDate = this.get("enablePreviousDate");
```
patrixr, line 39: maybe `enablePastDate` makes more sense than `enablePreviousDate` ? 🤔

PR #786, opened 2020-06-28
Adding image in the response

bharat619:
As in the title


app/components/variable-height-textarea.js
```diff
@@ -9,8 +9,13 @@ const remoteSearch = (authToken, cb) => {
   new AjaxPromise("/mentionable_users", "GET", authToken, {
```
patrixr, line 9: We really shouldn't have an Ajax request inside a reusable component. Pretty big code smell. Rather the users should be passed to the component.

bharat619: @patrixr ,
This wont be a code smell because,
1. This is a remote call, and not all the times a mentionable user might be needed. So only on the trigger of `@` the fetching is done
2. The fetching is very specific to the stock app. Also, the mentioning feature will now only be available for stock and admin users. So its very specific to this particular component itself.

PR #790, opened 2020-07-02
GCW-3189 User should not be able to change member of a set or a box/pallet to an invalid type

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3189
NOTE: Specify ticket link

### What does this PR do?
Fixes invalid package type selection for box/pallet/set

### Impacted Areas
Item detail screen

### DIFF
--- a/app/controllers/items/detail.js
+++ b/app/controllers/items/detail.js
@@ -84,6 +84,10 @@ export default GoodcityController.extend(
       return !!this.get("item.detail.length");
     },
 
+    packageSetName: Ember.computed("model", function() {
+      return this.get("model.packageSet.code.name");
+    }),
+
     showSetList: Ember.computed("_showSetList", "model.isPartOfSet", {
       get() {
         return this.get("_showSetList") && this.get("model.isPartOfSet");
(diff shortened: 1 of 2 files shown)


app/controllers/items/detail.js
```diff
@@ -570,9 +574,20 @@ export default GoodcityController.extend(
...
 
       async updatePackageType() {
-        const pkgType = await this.get(
-          "packageTypeService"
-        ).userPickPackageType();
+        let pkgType;
+        if (this.get("model.isPartOfSet")) {
+          const allowedPackageTypes = this.get("packageTypeService").parentsOf(
+            this.get("model.code")
+          );
```
patrixr, line 581: Shouldn't it be the children of the packageSet's type instead ?

app/controllers/items/detail.js
```diff
@@ -570,9 +574,28 @@ export default GoodcityController.extend(
...
+        let pkgType;
+
+        if (this.get("model.isPartOfSet")) {
+          const packageSet = this.get("model.packageSet");
+          const pkg = await this.get("packageService").userPickPackage({
+            packageTypes: this.get("packageService").allChildPackageTypes(
+              packageSet
+            ),
+            storageTypeName: "Package" // we don't add boxes to sets
+          });
```
patrixr, line 586: Why is it picking a package instead of a type ? 🤔 
The goal is to select a new type for the existing package, right ?

patrixr: I would imagine something like this :

```javascript
const types = this.get('packageTypeService');
const allowedPackageTypes = this.get("model.isPartOfSet") ?
             types.chidlrenTypesOf(this.get('model.packageSet.packageType')) : 
             types.getAllTypes();

const newType = await this.get("packageTypeService").userPickPackageType({ 
     subsetPackageTypes: allowedPackageTypes
});

// save the new package type
```

app/controllers/items/detail.js
```diff
@@ -570,9 +578,18 @@ export default GoodcityController.extend(
...
-        const pkgType = await this.get(
-          "packageTypeService"
-        ).userPickPackageType();
+        let pkgType;
+
+        if (this.get("model.isPartOfSet")) {
+          pkgType = await this.get("packageTypeService").userPickPackageType({
+            subsetPackageTypes: this.get(
+              "model.packageSet.code"
+            ).defaultChildPackagesList()
```
patrixr, line 587: Does the child package list include the type itself ?

bharat619: yup

PR #794, opened 2020-07-06
GCW-3185- packages history

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3185

### What does this PR do?

FEATURE:  On history tab in Item it shows comprehensive history of `Actions` and `version` changes applied on Item.

Screenshot 2020-07-02 at 11 49 21 PM
Screenshot 2020-07-02 at 11 49 36 PM

### DIFF
--- a/app/adapters/item_action.js
+++ b/app/adapters/item_action.js
@@ -1,9 +1,11 @@
 import ApplicationAdapter from "./application";
+import _ from "lodash";
 import config from "stock/config/environment";
 
 export default ApplicationAdapter.extend({
   urlForQuery(query, modelName) {
     const { NAMESPACE, API_HOST_URL } = config.APP;
-    return `${API_HOST_URL}/${NAMESPACE}/packages_inventories`;
+    const paramStr = _.map(query, (value, key) => `${key}=${value}`).join("&");
+    return `${API_HOST_URL}/${NAMESPACE}/packages_inventories?${paramStr}`;
   }
 });
--- a/app/adapters/item_action.js
+++ b/app/adapters/item_action.js
@@ -1,0 +1,9 @@
+import ApplicationAdapter from "./application";
+import config from "stock/config/environment";
+
+export default ApplicationAdapter.extend({
+  urlForQuery(query, modelName) {
+    const { NAMESPACE, API_HOST_URL } = config.APP;
+    return `${API_HOST_URL}/${NAMESPACE}/packages_inventories`;
+  }
+});
--- a/app/controllers/items/history.js
+++ b/app/controllers/items/history.js
@@ -15,11 +15,7 @@ export default detail.extend(SearchMixin, {
           this.getPaginationQuery(pageNo)
         )
       );
-      return this.get("store")
-        .query("item_action", params)
-        .then(results => {
-          return results;
-        });
+      return this.get("store").query("item_action", params);
     }
   }
 });
--- a/app/controllers/items/history.js
+++ b/app/controllers/items/history.js
@@ -1,0 +1,25 @@
+import detail from "./detail";
+import Ember from "ember";
+import _ from "lodash";
+import SearchMixin from "stock/mixins/search_resource";
+
+export default detail.extend(SearchMixin, {
+  autoLoad: true,
+  perPage: 25,
+
+  actions: {
+    loadMoreActions(pageNo) {
+      const params = this.trimQuery(
+        _.merge(
+          { package_id: this.get("model.id") },
+          this.getPaginationQuery(pageNo)
+        )
+      );
+      return this.get("store")
+        .query("item_action", params)
+        .then(results => {
+          return results;
+        });
+    }
+  }
+});
--- a/app/controllers/items/history.js
+++ b/app/controllers/items/history.js
@@ -31,18 +31,40 @@ export default detail.extend(SearchMixin, {
           if (!items.get("length")) {
             return [];
           }
-          let sortedItems = [...items.toArray()].sortBy("createdAt");
+          let versions = items
+            .get("firstObject")
+            .get("item.versions")
+            .toArray();
+          versions.shift();
+          let actionsAndVersions = [...items.toArray(), ...versions].sortBy(
+            "createdAt"
+          );
           const groups = _.reduce(
-            sortedItems,
+            actionsAndVersions,
             (results, action) => {
-              const groupKey = `${action.get("user.fullName")}/${action.get(
-                "action"
-              )}/${moment(action.get("createdAt")).format("LL")}`;
+              let groupKey;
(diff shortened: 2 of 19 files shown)


app/adapters/item_action.js
```diff
@@ -0,0 +1,11 @@
...
+import _ from "lodash";
+import config from "stock/config/environment";
+
+export default ApplicationAdapter.extend({
+  urlForQuery(query, modelName) {
+    const { NAMESPACE, API_HOST_URL } = config.APP;
+    const paramStr = _.map(query, (value, key) => `${key}=${value}`).join("&");
+    return `${API_HOST_URL}/${NAMESPACE}/packages_inventories?${paramStr}`;
+  }
+});
```
patrixr, line 11: We've never needed that before ? how come it's there ? 🤔

app/controllers/items/history.js
```diff
@@ -0,0 +1,74 @@
...
+        "dispatched_quanitity",
+        "id",
+        "detail_type_id",
+        "created_at",
+        "received_at"
+      ]);
+    });
+    versions.forEach((version, index) =>
+      version.set("objectChanges", results[index])
+    );
```
patrixr, line 25: Don't modify data, recreate a fresh set and return that instead (pure functions)

app/controllers/items/history.js
```diff
@@ -0,0 +1,74 @@
...
+    if (isAction) {
+      return `${action.get("user.fullName")}/${action.get(
+        "action"
+      )}/${createdAt}`;
+    } else {
+      return `${action.get("whodunnitName")}/${action.get(
+        "state"
+      )}/${createdAt}`;
+    }
+  },
```
patrixr, line 40: I'd suggest keeping the logic that checks the type of the record _inside_ the method. Makes the method "smarter" or "more magical" : )

```suggestion
  getGroupKey(change) {
    const isAction = change.get('constructor.modelName') === "item_action"; 
    const createdAt = moment(change.get("createdAt")).format("LL");
    const author = change.get(isAction ? "user.fullName" : "whodunnitName");
    const type = change.get(isAction ? "action" : "state");
    
    return `${author}/${type}/${createdAt}`;
  },
```

app/controllers/items/history.js
```diff
@@ -0,0 +1,74 @@
...
+        const isItemAction = !!action.get("user.fullName");
+        const groupKey = this.getGroupKey(isItemAction, action);
+
+        results[groupKey] = results[groupKey] || {
+          key: groupKey,
+          type: isItemAction ? action.get("action").capitalize() : "Edited",
+          date: createdAt,
+          user: isItemAction
+            ? action.get("user.fullName")
+            : action.get("whodunnitName"),
```
patrixr, line 56: i see you're also doing similar logic here. So might be handy to have some helper methods like `getAuthor(change)` which you can use everywhere in your code. Those could have the ternary inside, which would clear up a lot of the code.

Since you don't actually know whether it's an inventory record or a version record, using a variable named "change" makes a bit more sense to me (more generic). Just a personal preference though

app/controllers/items/history.js
```diff
@@ -0,0 +1,74 @@
...
+          actions: []
+        };
+        results[groupKey].actions.push(action);
+        return results;
+      },
+      {}
+    );
+    return _.values(groups);
```
patrixr, line 64: Probably want to sort by date as well

app/controllers/items/history.js
```diff
@@ -0,0 +1,74 @@
...
+  },
+
+  groupedActionsAndVersions: Ember.computed("model", "itemActions", function() {
+    let actionsAndVersions = [
+      ...this.get("itemActions").toArray(),
+      ...this.groupedVersions().toArray()
+    ].sortBy("createdAt");
+    return this.groupingAndSortingActions(actionsAndVersions);
```
patrixr, line 72: `groupingAndSortingActions` should also be sorting the actions (given its name).

The `.sortBy("createdAt");` should be moved into the method, _after_ the grouping is done (not before).

So something like `_.values(groups).sort(...)`

app/helpers/format-display-value.js
```diff
@@ -0,0 +1,31 @@
+import Ember from "ember";
+
+export default Ember.Helper.extend({
+  store: Ember.inject.service(),
+
+  getRecordName(modelName, modelId) {
```
patrixr, line 6: I'm not a big fan of having a helper lookup the store for records, but you can pass the model itself rather than the ID and do a `.get` directly on it. That'd be preferable.

Also, this looks way too specific to be a helper. Maybe this mapping should take effect when you build the normalized records (in the grouping code)

like

```
results[groupKey].actions.push(normalize(action));
```

app/helpers/startcase-string.js
```diff
@@ -0,0 +1,6 @@
+import Ember from "ember";
+import _ from "lodash";
+
+export default Ember.Helper.helper(function(value) {
+  return _.startCase(value[0]);
+});
```
patrixr, line 6: Isn't that the same as capitalize ?

app/models/item_action.js
```diff
@@ -26,5 +26,32 @@ export default DS.Model.extend({
...
+    return Number(this.get("quantity")) >= 0 ? "To" : "From";
+  }),
+
+  sourceSubscript: Ember.computed("quantity", function() {
+    return Number(this.get("quantity")) >= 0 ? "From" : "To";
+  }),
+
+  arrow: Ember.computed("quantity", function() {
+    return Number(this.get("quantity")) >= 0 ? "arrow-up" : "arrow-down";
+  }),
```
patrixr, line 45: Those probably shouldn’t be in the model, we should keep it data-centric. This feels more like UI related, maybe a helper for it ?

app/routes/items/history.js
```diff
@@ -1,15 +1,34 @@
...
 export default AuthorizeRoute.extend({
   model(params) {
-    return this.store.findRecord("item", params.item_id, { reload: true });
+    const queryParams = {
+      package_id: params.item_id,
+      per_page: 100
+    };
+    return Ember.RSVP.hash({
+      item: this.store.findRecord("item", params.item_id, { reload: true }),
+      itemActions: this.store.query("item_action", queryParams)
```
patrixr, line 12: @abulsayyad123 Could you investigate how much we're using the term `item_action` in the project ? I really think `inventory_action` makes a lot more sense. I'm hoping to refactor it

app/adapters/item_action.js
abulsayyad123: Yes.. I need to remove this.

app/helpers/startcase-string.js
abulsayyad123: This removes any separator like  underscore, hyphen etc and make it capitalize.

PR #803, opened 2020-07-20
GCW-3209 Stocktake UI

patrixr:
Ticket https://jira.crossroads.org.hk/browse/GCW-3209
Dependent on https://github.com/crossroads/api.goodcity/pull/1035

This PR adds a new section (accessible via the menu on the far right).

This section allows: 
- Creating/Editing/Cancelling stocktakes
- Adding quantity revisions to stocktakes (meaning a count of a package)
- Applying a stocktake to the inventory


app/helpers/array-find.js
```diff
@@ -0,0 +1,15 @@
+import Ember from "ember";
```
abulsayyad123, line 1: Nice!!

PR #805, opened 2020-07-20
GCW-3169 Stock-Item Messages Screen

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3169

### What does this PR do?
- Stock-Item Messages Screen
- Allow User to send messages on stock-items who has permission `can_manage_package_messages`
- For other users, display a message as `"You are not authorized to send messages on items."`
- List stock-item notifications in alerts tab.


app/mixins/messages_base.js
```diff
@@ -0,0 +1,130 @@
+import Ember from "ember";
```
bharat619, line 1: Awesome 💯

app/controllers/my_notifications.js
```diff
@@ -157,15 +149,20 @@ export default Ember.Controller.extend({
...
+        state: state,
+        messageable_type: ["order", "package"]
       };
 
-      return this.get("store")
-        .query("designation", params)
-        .then(orders => this.toOrderModels(orders.content))
-        .then(orders => {
-          const notifications = _.chain(orders)
+      return new AjaxPromise(
```
abulsayyad123, line 156: can we move this to service ?

app/models/item.js
```diff
@@ -390,6 +394,14 @@ export default cloudinaryUrl.extend(GradeMixin, {
     }
   ),
 
+  unreadMessagesCount: Ember.computed("messages.@each.state", function() {
+    return this.get("messages").filterBy("state", "unread").length;
+  }),
+
+  hasUnreadMessages: Ember.computed("unreadMessagesCount", function() {
```
abulsayyad123, line 401: Can we use `gt` computed property here ?

app/controllers/items/staff_conversation.js
```diff
@@ -0,0 +1,17 @@
...
+  isPrivate: true,
+
+  canManageItemMessages: Ember.computed.alias(
+    "session.currentUser.canManageItemMessages"
+  ),
+
+  actions: {
+    sendMessage() {
+      let values = this.prepareMessageObject("Package");
+      this.createMessage(values);
```
patrixr, line 14: There doesn't seem to be any error management here

app/controllers/my_notifications.js
```diff
@@ -32,25 +30,30 @@ export default Ember.Controller.extend({
...
     const store = this.get("store");
     const msg = store.peekRecord("message", notification.record.id);
-    const orderId = notification.record.order_id;
+    const messageableId = notification.record.messageable_id;
     const notifications = this.get("notifications");
 
-    if (!orderId || this.session.router.currentURL !== "/my_notifications") {
+    if (
+      !messageableId ||
+      this.session.router.currentURL !== "/my_notifications"
```
patrixr, line 38: Since this is the `my_notifications` controller, would the currentURL ever be something other than `/my_notifications` ?

app/controllers/my_notifications.js
```diff
@@ -101,17 +107,17 @@ export default Ember.Controller.extend({
   },
 
   /**
-   * Transform offers into "notifications" object with more UI-friendly properties
+   * Transform messages into "notifications" object with more UI-friendly properties
    *
-   * @param {Offer} offer
+   * @param {Meesage} messages
```
patrixr, line 112: small typo : )

app/controllers/my_notifications.js
patrixr: +1

app/mixins/messages_base.js
```diff
@@ -0,0 +1,130 @@
...
+        this.set("displayText", "");
+      })
+      .catch(error => {
+        this.store.unloadRecord(message);
+        throw error;
+      });
+  },
+
+  prepareMessageObject: function(messageableType) {
+    Ember.$("textarea").trigger("blur");
```
patrixr, line 65: Good idea creating a mixin for messages. However we can't really have JQuery code inside a mixin, for the simple reason that it should be reusable within any component/controller. So you can't assume that there is a `textarea` in the page.

That part can be kept in the controller : )

app/mixins/messages_base.js
```diff
@@ -0,0 +1,130 @@
...
+
+      let scrollOffset = Ember.$(document).height();
+      let screenHeight = document.documentElement.clientHeight;
+      let pageHeight = document.documentElement.scrollHeight;
+
+      if (pageHeight > screenHeight) {
+        Ember.run.later(this, function() {
+          window.scrollTo(0, scrollOffset);
+        });
+      }
```
patrixr, line 112: Same comment as above for this chunk of code, too page-specific to be in a mixin

app/controllers/items/staff_conversation.js
swatijadhav: Yes, it is handled here https://github.com/crossroads/stock.goodcity/pull/805/files#diff-ed05eabf206ef70eacdb8b04b22c229fR58

app/routes/items/staff_conversation.js
```diff
@@ -0,0 +1,38 @@
...
+        : []
+    });
+  },
+
+  setupController(controller, model) {
+    this._super(controller, model);
+    controller.set("messages", model.messages);
+    controller.set("model", model.item);
+
+    if (this.get("canManageItemMessages")) {
```
patrixr, line 27: `on` and `off` methods are meant to be lifecycle hooks. So they should always be called. But the conditional `this.get("canManageItemMessages")` can exist within those methods to do X or Y

app/controllers/items/staff_conversation.js
patrixr: But we would never know something went wrong here. Other than seeing an uncaught promise in the console (cause it throws the error).

You should probably `return` the promise, catch the error and display an error message. (you can use async mixin for that)

app/models/item.js
swatijadhav: `.gt` not available in Ember 2.15

app/mixins/messages_base.js
swatijadhav: This code is to mark the incoming messages in thread as read and scroll to bottom of page. Since this behavior is common on all message pages, its moved to mixin.

app/mixins/messages_base.js
```diff
@@ -0,0 +1,137 @@
...
+    this.get("messages").pushObject(message._internalModel);
+    this.get("messagesUtil").markRead(message);
+
+    this.scrollToBottom();
+  },
+
+  scrollToBottom() {
+    if (!Ember.$(".message-textbar").length) {
+      return;
+    }
```
patrixr, line 108: Since this is a mixin, we can't assume this element will be present. Maybe it's better to keep the scroll aspect of this in the controller

PR #817, opened 2020-07-23
GCW-3213-Order Administrator can now view dashboard

shreyas098:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3213

### What does this PR do?

BUG: Order Administrator can now view dashboard(home page) in stock app

### DIFF
--- a/app/models/user_profile.js
+++ b/app/models/user_profile.js
@@ -41,11 +41,22 @@ export default Addressable.extend({
     }
   }),
 
+  isOrderAdministrator: Ember.computed("roleNames", function() {
+    if (this.get("roleNames") && this.get("roleNames").length) {
+      return this.get("roleNames").indexOf("Order administrator") >= 0;
+    }
+  }),
+
   canViewDashboard: Ember.computed(
     "isOrderFulfilmentUser",
     "isSupervisor",
+    "isOrderAdministrator",
     function() {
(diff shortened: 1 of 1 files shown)


app/models/user_profile.js
```diff
@@ -41,11 +41,22 @@ export default Addressable.extend({
     }
   }),
 
+  isOrderAdministrator: Ember.computed("roleNames", function() {
+    if (this.get("roleNames") && this.get("roleNames").length) {
+      return this.get("roleNames").indexOf("Order administrator") >= 0;
```
abulsayyad123, line 46: I know you have done as per previous code but this `Order administrator` hardcode doesn't seems correct. A single change of role name on api will break the app.
Just, a thought is it fine to preload the permissions on the app @patrixr ?
cc: @bharat619 @swatijadhav

bharat619: I really like the idea. We are fetching the `role_permissions`, we can extend it to include permission names. We can remove a lot of dependency of role based check made in methods like `canViewDashboard` 
I think this goes very much in hand of what @steveyken was suggesting to slowly move away from Role name and stick with permissions

abulsayyad123: If this is in priority, we can go with this solution for now. @shreyas098, can you please create ticket for refactoring the hard coding of permissions on stock app and in description of this ticket add this conversation.

PR #825, opened 2020-07-29
GCW-3230- Adding subform versioning 

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3230

### What does this PR do?

FEATURE: This PR displays subforms versions along with package versions in package history page.

### DIFF
--- a/app/helpers/format-display-value.js
+++ b/app/helpers/format-display-value.js
@@ -6,13 +6,13 @@ export default Ember.Helper.extend({
   store: Ember.inject.service(),
   i18n: Ember.inject.service(),
 
-  getRecordName(modelName, modelId) {
+  getRecordName(modelName, modelId, fieldName = "name") {
     if (!modelId) {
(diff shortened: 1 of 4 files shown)


app/helpers/format-display-value.js
```diff
@@ -43,6 +63,16 @@ export default Ember.Helper.extend({
...
       case "saleable":
         return this.getSaleable(value);
+      case "frequency_id":
+        return this.getLookUp(value);
+      case "voltage_id":
+        return this.getLookUp(value);
+      case "test_status_id":
+        return this.getLookUp(value);
+      case "comp_test_status_id":
+        return this.getLookUp(value);
```
patrixr, line 73:
```suggestion
      case "frequency_id":
      case "voltage_id":
      case "test_status_id":
      case "comp_test_status_id":
        return this.getLookUp(value);
```

PR #848, opened 2020-08-19
GCW-3276- Multiple favorite printers

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3276

### What does this PR do?

This PR adds new endpoint for updating and creating favourite printer for user in stock.

### DIFF
--- a/app/routes/items/new.js
+++ b/app/routes/items/new.js
@@ -50,15 +50,15 @@ export default AuthorizeRoute.extend(GradeMixin, {
   },
 
   setupPrinterId(controller) {
-    let allAvailablePrinters = this.get(
-      "printerService"
-    ).allAvailablePrinters();
-    let user = this.get("session.currentUser");
-    if (user.get("printerId")) {
-      controller.set("selectedPrinterId", user.get("printerId"));
+    let defaultPrinter = this.get("printerService").getDefaultPrinter();
+    if (defaultPrinter) {
+      controller.set("selectedPrinterId", defaultPrinter.id);
     } else {
+      let allAvailablePrinters = this.get(
+        "printerService"
+      ).allAvailablePrinters();
       let firstPrinterId = allAvailablePrinters[0].id;
-      this.get("printerService").updateUserDefaultPrinter(firstPrinterId);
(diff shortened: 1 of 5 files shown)


app/routes/items/new.js
```diff
@@ -50,15 +50,15 @@ export default AuthorizeRoute.extend(GradeMixin, {
...
-    if (user.get("printerId")) {
-      controller.set("selectedPrinterId", user.get("printerId"));
+    let defaultPrinter = this.get("printerService").getDefaultPrinter();
+    if (defaultPrinter) {
+      controller.set("selectedPrinterId", defaultPrinter.id);
     } else {
+      let allAvailablePrinters = this.get(
+        "printerService"
+      ).allAvailablePrinters();
       let firstPrinterId = allAvailablePrinters[0].id;
```
patrixr, line 60: This line would break if `allAvailablePrinters` is an empty array

abulsayyad123: ohh !! thanks for pointing, will add guard clause.

app/services/printer-service.js
```diff
@@ -11,12 +12,38 @@ export default Ember.Service.extend({
...
   },
+
   updateUserDefaultPrinter(printerId) {
-    new AjaxPromise(
-      `/users/${this.get("session.currentUser.id")}`,
-      "PUT",
-      this.get("session.authToken"),
-      { user: { printer_id: printerId } }
-    ).then(data => this.get("store").pushPayload(data));
+    const defaultPrinterUser = this.__getStockPrintersUsers().get(
```
patrixr, line 17: The default printer record might not exist if it's a new user, so this would break.
Maybe it's better to do a "create if missing" behaviour ?

abulsayyad123: Sure.. i'll add it. Thanks

abulsayyad123: We are assigning the first printer  as the default printer if there isn't any for new user.

patrixr (approved):

bharat619 (approved):

swatijadhav (approved):

bharat619 (approved):

PR #852, opened 2020-08-24
GCW-2957-Added fix for dropbox in management quota

shreyas098:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2957

### What does this PR do?
On the 'Manage Appointment Quotas' page, after selecting a 3 vertical dots icon on a day and opening the dropdown box, if a user goes to the 'Special Dates' tab and returns to the 'Default Quotas' tab, the dropdown box will still be open fix

### DIFF
--- a/app/controllers/appointments.js
+++ b/app/controllers/appointments.js
@@ -130,10 +130,20 @@ export default GoodcityController.extend({
 
     showSpecial() {
       this.set("selectedTab", "special");
+      _.map(this.get("specialSlotsByDate"), days => {
+        if (days.get("showOptionsMenu")) {
+          days.set("showOptionsMenu", !days.get("showOptionsMenu"));
+        }
+      });
     },
 
     showDefault() {
       this.set("selectedTab", "presets");
+      _.map(this.get("presetsByWeekDay"), days => {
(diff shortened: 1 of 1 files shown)


app/controllers/appointments.js
```diff
@@ -130,10 +130,20 @@ export default GoodcityController.extend({
...
+      });
     },
 
     showDefault() {
       this.set("selectedTab", "presets");
+      _.map(this.get("presetsByWeekDay"), days => {
+        if (days.get("showOptionsMenu")) {
+          days.set("showOptionsMenu", !days.get("showOptionsMenu"));
+        }
+      });
```
patrixr, line 146: `map` is used to transform an array by returning a new element for each item, if you don't return anything inside the predicate, then you should use `each` or `forEach` instead

```suggestion
      _.each(this.get("presetsByWeekDay"), days => days.set("showOptionsMenu", false))
```

or

```suggestion
      this.get("presetsByWeekDay").forEach(days => days.set("showOptionsMenu", false))
```

abulsayyad123 (approved):

bharat619 (approved):

PR #854, opened 2020-08-26
GCW-3247 Fix Unexpected Error Message

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3247

### What does this PR do?
Fixes the error message while creating item.

### DIFF
--- a/app/routes/items/new.js
+++ b/app/routes/items/new.js
@@ -53,7 +53,7 @@ export default AuthorizeRoute.extend(GradeMixin, {
     let allAvailablePrinters = this.get(
       "printerService"
     ).allAvailablePrinters();
-    let user = this.get("session.loggedInUser");
+    let user = this.get("session.currentUser");
     if (user.get("printerId")) {
       controller.set("selectedPrinterId", user.get("printerId"));
     } else {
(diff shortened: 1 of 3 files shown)


app/routes/items/new.js
```diff
@@ -162,6 +162,5 @@ export default AuthorizeRoute.extend(GradeMixin, {
     controller.set("defaultCondition", this.getDefaultCondition());
     controller.set("valueHkDollar", +defaultValue.value_hk_dollar);
     controller.set("defaultValueHkDollar", +defaultValue.value_hk_dollar);
-    this.setupPrinterId(controller);
```
bharat619, line 165: removed this as it is also called [here](https://github.com/crossroads/stock.goodcity/pull/854/files#diff-582238b7d4945ca1309bf7b22d97b16bR74)

config/environment.js
```diff
@@ -94,7 +94,6 @@ module.exports = function(environment) {
         "packages_location",
         "stockit_contact",
         "goodcity_request",
-        "donor_condition",
```
abulsayyad123, line 97: why removed donor_condition ?

patrixr (approved):

config/environment.js
patrixr: same question

bharat619: because the donor_condition is being unloaded from store. And it dosn't actually needs to be unloaded

swatijadhav (approved):

abulsayyad123 (approved):

shreyas098 (approved):

PR #855, opened 2020-08-27
GCW-3280-Add new user feature

shreyas098:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3280

### What does this PR do?

FEATURE: Adds create new user screen and contact details screen for new user

### Screenshots:

Screen Shot 2020-09-13 at 3 02 00 AM
Screen Shot 2020-09-13 at 3 03 59 AM
Screen Shot 2020-09-13 at 3 03 42 AM


app/controllers/users/add_user.js
```diff
@@ -0,0 +1,148 @@
...
+      { name: ms, id: "Ms" }
+    ];
+  }),
+
+  setEmailorMobile: Ember.computed("email", "mobilePhone", function() {
+    const email = this.get("email");
+    const mobile = this.get("mobilePhone");
+
+    if (/^[456789]\d{7}/.test(mobile) || /^[^@\s]+@[^@\s]+/.test(email)) {
+      return false;
```
bharat619, line 32: by reading the method name, this should set either email or mobile, but its actually setting boolean value whose value is used in required field.
This method name needs to be changed

app/controllers/users/add_user.js
```diff
@@ -0,0 +1,148 @@
...
+
+  formatMobileNumber() {
+    const mobile = this.get("mobilePhone");
+    if (mobile.length) {
+      return config.APP.HK_COUNTRY_CODE + mobile;
+    }
+  },
+
+  getRequestParams() {
+    const mobilePhone = this.get("mobilePhone")
```
bharat619, line 75: `mobileNumber` gives better readability, instead of `mobilePhone`

app/controllers/users/add_user.js
```diff
@@ -0,0 +1,148 @@
...
+    },
+
+    saveUser() {
+      return this.runTask(async () => {
+        let data = await new AjaxPromise(
+          "/users",
+          "POST",
+          this.get("session.authToken"),
+          this.getRequestParams()
+        );
```
bharat619, line 120: the part where a request is fired, can be put in a user-service

```suggestion
        let data = await this.get('userService').create(this.getRequestParams())
```
and the userService should handle the request and authToken thing

app/controllers/users/add_user.js
```diff
@@ -0,0 +1,148 @@
...
+        this.get("store").pushPayload(data);
+        let newUser = this.get("store").peekRecord("user", data.user.id);
+        newUser.set("image", await this.saveImage());
+        const { id } = await newUser.save();
```
bharat619, line 124: What you are doing is
1. creating a new user record
2. creating a new image record
3. update the user record with the image

Probably you can rewrite the function to do it as:
1. create image record, if image exists
2. create user record with image

Avoids extra update network request

app/controllers/users/add_user.js
```diff
@@ -0,0 +1,148 @@
...
+            this,
+            function() {
+              this.transitionToRoute("manage_users");
+              this.clearFormData();
+            },
+            0
+          );
+        },
+        this.locale("no").string
+      );
```
bharat619, line 145: You can use the `modalConfirm` method from async mixing

```suggestion
      const confirmed = await this.modalConfirm(
        this.get("i18n").t("users.cancel_user_warning")
      );

      if (!confirmed) {
        return;
      }
      
      this.transitionToRoute("manage_users");
      this.clearFormData();
```

app/controllers/users/contact_details.js
```diff
@@ -0,0 +1,174 @@
...
+    let miss = translation.t("order.user_title.miss");
+    let ms = translation.t("order.user_title.ms");
+
+    return [
+      { name: mr, id: "Mr" },
+      { name: mrs, id: "Mrs" },
+      { name: miss, id: "Miss" },
+      { name: ms, id: "Ms" }
+    ];
+  }),
```
bharat619, line 22: I smell code duplication here. This was used in addUser.js as well

Probably you can make a method in userService, which you can reuse in both

app/controllers/users/contact_details.js
```diff
@@ -0,0 +1,174 @@
...
+
+  languages: Ember.computed(function() {
+    let translation = this.get("i18n");
```
bharat619, line 25: Probably you can make a method in userService, which you can reuse in both

app/controllers/users/contact_details.js
```diff
@@ -0,0 +1,174 @@
...
+    let address_attributes = {
+      district_id: this.get("user.associatedDistrict.id") || null
+    };
+    return { address_attributes };
+  },
+
+  saveImage() {
+    let image = this.get("newUploadedImage");
+    if (image) {
+      return image.save();
```
bharat619, line 58: since this would fire an request, probably you can do some error handling as well

app/controllers/users/contact_details.js
```diff
@@ -0,0 +1,174 @@
...
+    }
+  },
+  deleteOldImage(img) {
+    if (img) {
+      img.deleteRecord();
```
bharat619, line 63: since this would fire an request, probably you can do some error handling as well

app/controllers/users/contact_details.js
```diff
@@ -0,0 +1,174 @@
...
+      return img.save();
+    }
+  },
+  actions: {
+    focusIn(field, value) {
+      this.set(`${field}previousValue`, value);
```
bharat619, line 69: whats the intent of this method? 🤔 
Are you setting the previousValue on focus in?

If so, then you can remove this, and probably make use of ember `dirty` methods, which already keeps track of changed attributes and its values

app/controllers/users/contact_details.js
```diff
@@ -0,0 +1,174 @@
...
+    },
+
+    focusOut(field, value, required) {
```
bharat619, line 72: This method needs a lot of refactoring. Have a look of how `app/controllers/organisation/detail.js` is done

app/controllers/users/contact_details.js
```diff
@@ -0,0 +1,174 @@
...
+        newUser.set("image", await this.saveImage());
+        newUser.save();
+      }, ERROR_STRATEGIES.MODAL);
+    },
+
+    deleteImage() {
+      this.runTask(async () => {
+        let newUser = this.get("store").peekRecord("user", this.get("user.id"));
+        await this.deleteOldImage(newUser.get("image"));
+        let data = await new AjaxPromise(
```
bharat619, line 151: as mentioned above, move this to a separate service

app/templates/users/add_user.hbs
```diff
@@ -0,0 +1,147 @@
...
+    </div>
+  </div>
+
+  <div class="row">
+    <div class="small-5 columns item-label">
+      {{t 'users.email'}}
+    </div>
+    <div class="small-7 columns field-container">
+      {{#form-control}}
+        {{input value=email name="email" id="email" type="email" required=setEmailorMobile }}
```
bharat619, line 78: As i mentioned above, any person reading this would get confused
`required=setEmailorMobile`
always important to have a proper naming convention : )

bharat619 (changes requested):

app/controllers/users/add_user.js
```diff
@@ -1,166 +1,121 @@
...
+    isEmailorMobilePresent: Ember.computed("email", "mobileNumber", function() {
+      const email = this.get("email");
+      const mobile = this.get("mobileNumber");
+
+      if (/^[456789]\d{7}/.test(mobile) || /^[^@\s]+@[^@\s]+/.test(email)) {
+        return false;
       } else {
-        this.get("selectedRoleIds").removeObject(id);
+        return true;
       }
```
swatijadhav, line 25:
```js
return !Boolean( /^[456789]\d{7}/.test(mobile) || /^[^@\s]+@[^@\s]+/.test(email) ) 
```

app/controllers/users/add_user.js
```diff
@@ -1,166 +1,121 @@
...
-            0
+      saveUser() {
+        return this.runTask(async () => {
+          let data = await this.get("userService").createUser(
+            await this.getRequestParams()
           );
-        },
-        this.locale("no").string
-      );
+          console.log(data);
```
swatijadhav, line 102:
```suggestion
```

app/controllers/users/contact_details.js
```diff
@@ -0,0 +1,165 @@
...
+
+export default Ember.Controller.extend(
+  AsyncMixin,
+  ImageUploadMixin,
+  TitleAndLanguageMixin,
+  {
+    user: Ember.computed.alias("model.user"),
+    userService: Ember.inject.service(),
+
+    emailIdNotPresent: Ember.computed("user.email", function() {
```
swatijadhav, line 15: Can we better rename `emailIdNotPresent` to `invalidEmail` ?

app/controllers/users/contact_details.js
```diff
@@ -0,0 +1,165 @@
...
+      let email = this.get("user.email");
+      return !/^[^@\s]+@[^@\s]+/.test(email);
+    }),
+
+    mobileNotPresent: Ember.computed(
```
swatijadhav, line 20: Can we better rename `mobileNotPresent ` to `invalidMobile` ?

app/controllers/users/contact_details.js
```diff
@@ -0,0 +1,165 @@
...
+          newUser.save();
+        }, ERROR_STRATEGIES.MODAL);
+      },
+
+      deleteImage() {
+        this.runTask(async () => {
+          let newUser = this.get("store").peekRecord(
+            "user",
+            this.get("user.id")
+          );
```
swatijadhav, line 144: we can use `user` value defines above (line 12)
https://github.com/crossroads/stock.goodcity/pull/855/files#diff-c94f26135328a1f7cc107818ba4d2aa7R12

swatijadhav:
```suggestion
          let user = this.get("user");
```

app/controllers/users/contact_details.js
```diff
@@ -0,0 +1,165 @@
...
+  AsyncMixin,
+  ImageUploadMixin,
+  TitleAndLanguageMixin,
+  {
+    user: Ember.computed.alias("model.user"),
+    userService: Ember.inject.service(),
+
+    emailIdNotPresent: Ember.computed("user.email", function() {
+      let email = this.get("user.email");
+      return !/^[^@\s]+@[^@\s]+/.test(email);
```
bharat619, line 17: move the regex to constant. As this has been used over 4 different places

app/controllers/users/contact_details.js
```diff
@@ -0,0 +1,165 @@
...
+
+    districts: Ember.computed(function() {
+      return this.get("store")
+        .peekAll("district")
+        .sortBy("name");
+    }),
+
+    getUpdatedDistrictParams() {
+      let address_attributes = {
+        district_id: this.get("user.associatedDistrict.id") || null
```
bharat619, line 36:
```suggestion
        district_id: this.get("user.associatedDistrict.id")
```

app/controllers/users/contact_details.js
```diff
@@ -0,0 +1,165 @@
...
+      return { address_attributes };
+    },
+
+    saveImage() {
+      let image = this.get("newUploadedImage");
+      if (image) {
+        return image.save();
+      }
+    },
+    deleteOldImage(img) {
```
bharat619, line 47: this is just deleting image if image its present or not. 
rename this to `deleteImage(img)`

and the logic of checking weather image is present or not is better to be performed from the place where its called.
with that, this method will have a single responsibility of only deleting the image

See if you can move it to service

app/controllers/users/contact_details.js
```diff
@@ -0,0 +1,165 @@
...
+
+    getUpdatedDistrictParams() {
+      let address_attributes = {
+        district_id: this.get("user.associatedDistrict.id") || null
+      };
+      return { address_attributes };
+    },
+
+    saveImage() {
+      let image = this.get("newUploadedImage");
```
bharat619, line 42: this method should be designed as the name suggests saveImage, meaning ONLY to save the image.

```
saveImage(img){
  image.save()
}
```

I would suggest to make the image creation/deletion as a service as I am seeing ur using it in multiple places

```javascript
export default ApiBaseService.extend({
   create(img) {
      // logic
  },
 
 delete(img) {
    // logic
 }
})

```

app/controllers/users/contact_details.js
```diff
@@ -0,0 +1,165 @@
...
+    },
+    deleteOldImage(img) {
+      if (img) {
+        img.deleteRecord();
+        return img.save();
+      }
+    },
+    actions: {
+      focusIn(field, value) {
+        this.set(`${field}previousValue`, value);
```
bharat619, line 55: i really dont like the idea of setting a value on focusIn and then checking if its been changed.

I would suggest remove this logic and make use of `this.get('model').changedAttributes()`

bharat619 (changes requested):

app/controllers/users/contact_details.js
shreyas098: sure!! seems correct to have it in service

app/controllers/users/contact_details.js
```diff
@@ -0,0 +1,149 @@
...
+        }
+      },
+
+      updateMobile(e) {
+        let value = e.target.value.trim();
+        if (value) {
+          this.set("user.mobile", "+852" + this.get("mobileNumber"));
+        } else {
+          this.set("user.mobile", this.get("mobileNumber"));
+        }
```
swatijadhav, line 91:
```suggestion
       let mobileCode = value ? "+852" : "";
       this.set("user.mobile", mobileCode + this.get("mobileNumber"));
```

app/mixins/grades_option.js
```diff
@@ -13,6 +13,29 @@ export default Ember.Mixin.create({
...
+      { name: ms, id: "Ms" }
+    ];
+  }),
+
+  languages: Ember.computed(function() {
+    let translation = this.get("i18n");
+    let English = translation.t("users.languages.english");
+    let Chinese = translation.t("users.languages.chinese");
+
+    return [{ name: English, id: "en" }, { name: Chinese, id: "zh-tw" }];
```
swatijadhav, line 36:
```suggestion
    let english = translation.t("users.languages.english");
    let chinese = translation.t("users.languages.chinese");

    return [{ name: english, id: "en" }, { name: chinese, id: "zh-tw" }];
```

swatijadhav (approved):

bharat619 (approved):

PR #857, opened 2020-09-01
GCW-3284 Add Create/Update Organisation UI

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3284

### What does this PR do?
Adds organisation create and update feature

### Screenshot
image

image


app/controllers/organisations/new.js
```diff
@@ -0,0 +1,128 @@
...
+          this.replaceRoute("/");
+        });
+      };
+
+      this.get("messageBox").custom(
+        this.get("i18n").t("organisation.cancel_warning"),
+        "Yes",
+        cancel,
+        "No"
+      );
```
patrixr, line 107: FYI: I added a helper in the async mixin to create confirmation modals which supports promises

e.g

```
const confirmed = await this.modalConfirm('organisation.cancel_warning');

if (confirmed) { something }
```

Note a change request, just worth mentioning : )

app/routes/organisations/new.js
```diff
@@ -0,0 +1,41 @@
...
+    });
+    const countryObj = country.get("firstObject");
+    this.store.pushPayload(country);
+    this.controller.set("country", {
+      id: countryObj.get("id"),
+      nameEn: countryObj.get("nameEn")
+    });
+    this.controller.set("countryValue", {
+      country_id: countryObj.get("id")
+    });
```
patrixr, line 34: Is `countryValue` needed if `country` already exists ?

app/routes/organisation.js
```diff
@@ -1,14 +1,62 @@
...
+    const selectedOrganisationType = this.store.peekRecord(
+      "organisationType",
+      model.get("organisationTypeId")
+    );
+    this.controller.set("selectedOrganisationType", selectedOrganisationType);
+  },
+
+  async getOrganisationTypes() {
+    const organisationTypes = this.store.peekAll("organisationTypes");
+    if (organisationTypes.content.length) {
```
patrixr, line 54:
```suggestion
    if (organisationTypes.get('length')) {
```

app/routes/organisation.js
```diff
@@ -1,14 +1,62 @@
...
+    this.controller.set("countryValue", { country_id: country.get("id") });
+  },
+
+  // set organisation users
+  async setOrganisationUsers(model) {
+    let gcOrganisationUsers = this.store
+      .peekAll("organisationsUser")
+      .filterBy("organisationId", parseInt(model.id));
+    this.controller.set("gcOrganisationUsers", gcOrganisationUsers);
+  },
```
patrixr, line 37: I'm trying to understand where the organisationsUsers are used within this feature 🤔  
cc @bharat619

app/styles/templates/organisations/_new.scss
```diff
@@ -0,0 +1,149 @@
+$dropdown-body: #000000;
+$input-border: #8091a9;
+$white: #fff;
+$background-color: #002352;
+$error-background: #f46769;
+
+.header-container {
```
patrixr, line 7: Can you please scope all the classes to to this page. We can't keep putting classes at the top level, it creates too many conflicts across the project

app/controllers/organisations/detail.js
```diff
@@ -1,6 +1,94 @@
...
+        }
+      });
+
+      if (
+        !this.get("name_en_error") &&
+        !this.get("country_error") &&
+        !this.get("website_error")
+      ) {
+        this.send("update", this.get("model"));
+      }
```
patrixr, line 49: Shouldn't you turn on "validate" every time ? 🤔

patrixr: Btw, I recommend not having the error properties depend on the `"validate"` flag. They should always return the error if present. 

The reason for that is that some other developer might come in one day and call `this.get("name_en_error")` with the expectation that he/she will get the true error state. Which could lead to some drama : )

Maybe you can turn on a `"showErrors`" property whenever you want the UI to show the errors

app/styles/templates/organisations/_new.scss
patrixr: this applies to all CSS files btw

patrixr (changes requested):
Added a few minor comments, looks good otherwise 😄

app/routes/organisation.js
bharat619: yes @patrixr , you are right.
its not a part of this feature. I just moved the existing organisation-users related code to a separate method

app/controllers/organisations/new.js
bharat619: done : )

app/controllers/organisations/detail.js
bharat619: done : )

app/routes/organisations/new.js
bharat619: oops. missed it. removed

app/styles/templates/organisations/_search_organisations.scss
```diff
@@ -1,7 +1,34 @@
+.organisation_search_container{
```
patrixr, line 1: @bharat619 Could you also fix the scoping here ?

I think we can probably move `.search-box-container` and `.search-box` into `.organisation_search_container`. 
`.search-box` is very generic and it can easily conflict with another element somewhere else in the app

I assume we can also put all of `.organisations_search` into `.organisation_search_container`

patrixr (approved):
Looking good 👍

app/controllers/organisations/new.js
```diff
@@ -0,0 +1,116 @@
...
+  messageBox: Ember.inject.service(),
+  i18n: Ember.inject.service(),
+
+  showError: false,
+  name_en: "",
+  name_zh_tw: "",
+  website: "",
+  validate: false,
+
+  isInValidNameEn: Ember.computed("name_en", function() {
```
abulsayyad123, line 18: I think we can move some of the common methods to mixins with name `organisations_validations` mixins ?
I have seen this common pattern of too much validations in `controller` which is making controlling too much bloated. I think we can create `mixins/validations` and use it in controllers.
Thoughts ?

@patrixr

patrixr: If you create a mixin just for organisations, then it likely won't ever be used anywhere else. Which defeats the purpose of a mixin.

The computed property `isInValidNameEn` with its own specific validation doesn't have a place in a mixin, because its name is custom to this controller, and it's validation logic is custom as well.

You can create a general validation mixin, e.g `mixins/validations` in which you can have helper methods for validations.

e.g maybe something like

```javascript
import ValidationMixin, { notEmpty, gte, isUrl, matches } from '../mixins/validations'

export default Controller.extend(ValidationMixin, {
    validations: {
        name:      [ notEmpty() ],
        website:   [ isUrl(), matches('goodcity.hk') ],
        age:       [ gte(18) ]
    }

   // The mixin code would read the configuration above ☝️  and generate computed properties like `isNameValid`
})
```

Basically --> you can put reusable tools in a mixin, even generated computed properties. But the controller must decide what/how it is structured

patrixr: I'm very much in support of writing reusable tools, but I think it should probably be done outside of the scope of this PR

app/routes/organisation.js
```diff
@@ -1,14 +1,57 @@
...
 
 export default AuthorizeRoute.extend({
   model(params) {
-    return this.store.peekRecord("gc_organisation", params.organisation_id) || 
-      this.store.findRecord("gc_organisation", params.organisation_id, { reload: true });
+    return (
+      this.store.peekRecord("gc_organisation", params.organisation_id) ||
+      this.store.findRecord("gc_organisation", params.organisation_id, {
+        reload: true
+      })
```
swatijadhav, line 9:
```suggestion
    return this.loadIfAbsent("gc_organisation", params.organisation_id);
```

app/controllers/organisations/new.js
```diff
@@ -0,0 +1,116 @@
...
+     *      type is present
+     *      website has a valid format iff its present
+     */
+    createOrganisation() {
+      if (
+        !this.get("isInValidNameEn") &&
+        !this.get("isInValidCountry") &&
+        !this.get("isInValidWebsite")
+      ) {
+        this.showLoadingSpinner();
```
abulsayyad123, line 48: I think you should've used runtask.

abulsayyad123 (approved):
One minor change related to using runTask which you add later.

swatijadhav (approved):

PR #869, opened 2020-09-09
GCW 3281 Add Charity Position

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3281

### What does this PR do?
- Adds a new page to add/update organisations_users
- Refactor search_organisation. Remove the search_organisation controllers and routes and make it as a overlay component

### Screenshots
image


app/controllers/organisations/detail.js
```diff
@@ -2,5 +2,15 @@ import Ember from "ember";
...
-  backLinkPath: ""
+  backLinkPath: "",
+  organisationService: Ember.inject.service(),
+
+  actions: {
+    async searchOrganisation() {
+      const organisation = await this.get(
+        "organisationService"
+      ).userPickOrganisation();
+      this.replaceRoute("organisations.detail", organisation.get("id"));
```
patrixr, line 13: I'm seeing this in a few different files, maybe you could create a mixin with some navigation presets ? 

```suggestion
    async navigateToOrganisation(org = null) {
      const organisation = org || await this.get(
        "organisationService"
      ).userPickOrganisation();
      this.replaceRoute("organisations.detail", organisation.get("id"));
```

Also, using an optional argument like this ☝️  would make it so that whoever uses your mixin has the option of pre-selecting the organisation. 
In the context of a navigation mixin, I think a name like `navigateToOrganisation` is less likely to conflict with custom actions from controllers

app/controllers/users/add_user.js
```diff
@@ -57,17 +57,6 @@ export default Ember.Controller.extend(AsyncMixin, ImageUploadMixin, {
...
-    if (this.get("organisationId")) {
-      let store = this.get("store");
-      let organisation = store.peekRecord(
-        "gc_organisation",
-        this.get("organisationId")
-      );
-      this.set("organisationName", organisation && organisation.get("nameEn"));
-    }
-  }),
-
```
patrixr, line 70: 👍  nice ! the fewer observers the better

app/controllers/users/charity_position.js
```diff
@@ -0,0 +1,90 @@
...
+      const organisationUser = this.get(
+        "organisationsUserService"
+      ).getOrganisationUser(organisation.get("id"), this.get("user_id"));
+
+      if (organisationUser) {
+        this.transitionToRoute(
+          `/users/${this.get(
+            "user_id"
+          )}/charity_position?id=${organisationUser.get("id")}`
+        );
```
patrixr, line 45: anyone else thinking we should remove that automatic prettifier ? I feel we should be allowed to pass in a long string to a `transition` method.

cc @bharat619 @steveyken @swatijadhav @abulsayyad123 @shreyas098

app/controllers/users/charity_position.js
```diff
@@ -0,0 +1,90 @@
...
+      );
+      this.set("model.status", this.get("selectedStatus.name"));
+      try {
+        this.runTask(async () => {
+          await this.get("model").save();
+          this.replaceRoute("users.details", this.get("user_id"));
+        }, ERROR_STRATEGIES.MODAL);
+      } catch (error) {
+        throw error;
+      }
```
patrixr, line 75: If you want the exception to propagate, you can use `ERROR_STRATEGIES.RAISE` instead

Also not that `runTask` returns a promise, so you'd have to `await` it for the try/catch to work here.

app/models/organisations_user.js
```diff
@@ -13,6 +14,11 @@ export default Model.extend({
   organisation: belongsTo("organisation", {
     async: false
   }),
+  user_id: attr("string"),
+  status: attr("string"),
```
patrixr, line 18: I think this branch deserves a rebase because I already added the `status` field in that file 🤔

app/routes/users/charity_position.js
```diff
@@ -0,0 +1,98 @@
...
+      const data =
+        model.get("organisation") ||
+        this.store.peekRecord("gc_organisation", this.get("organisation_id"));
+      controller.set("organisation", data);
+
+      const orgData =
+        this.store.peekRecord("organisation", data.get("id")) ||
+        this.store.createRecord("organisation", {
+          nameEn: data.get("nameEn"),
+          id: data.get("id")
```
patrixr, line 77: if there's an `id` shouldn't we `findRecord` instead ??

patrixr: actually, nevermind that, why are we using the `organisations` model at all ? shouldn't we stick to `gc_organisations` ? (which in reality is the real `organisations` on the api side ... 🤦 )

app/components/goodcity/organisation-search-overlay.js
```diff
@@ -1,10 +1,13 @@
 import Ember from "ember";
+import _ from "lodash";
+import SearchMixin from "stock/mixins/search_resource";
 
-export default Ember.Controller.extend({
+export default Ember.Component.extend(SearchMixin, {
```
patrixr, line 5: I think you pasted `Component` by mistake here : )

app/styles/app.scss
```diff
@@ -54,6 +54,7 @@
 @import "templates/organisations/organisation_tabs";
 @import "templates/organisations/search_organisations";
 @import "templates/users/new";
+@import "templates/users//charity_position";
```
shreyas098, line 57: Typo error

app/routes/users/charity_position.js
```diff
@@ -0,0 +1,98 @@
...
+        (await this.store.findRecord("organisations_user", params.id, {
+          reload: true
+        }))
+      );
+    } else {
+      return this.store.createRecord("organisations_user", {
+        position: "",
+        preferredContactNumber: "",
+        organisationId: "",
+        user_id: "",
```
patrixr, line 28:
```suggestion
        user_id: this.get('user_id'),
```

just guessing

app/controllers/users/details.js
```diff
@@ -0,0 +1,57 @@
+import Ember from "ember";
+
+export default Ember.Controller.extend({
+  organisationService: Ember.inject.service(),
+  organisationsUserService: Ember.inject.service(),
+  userOrganisationDetails: Ember.computed("model", function() {
```
patrixr, line 6:
```suggestion
  userOrganisationDetails: Ember.computed("model", "model.organisationsUsers.[]", "model.organisationsUsers.@each.userStatus", function() {
```

I fear it might not react properly otherwise

app/components/goodcity/organisation-search-overlay.js
bharat619: no 🤔 
This should be `Component` as its an overlay component 🤔

app/models/organisations_user.js
bharat619: got this after rebasing

app/routes/users/charity_position.js
bharat619: can do that. Its just an initialisation done here

Setting of the user is done in the controller

app/components/goodcity/organisation-search-overlay.js
patrixr: Oh, I didn't notice you moved it to another folder ! Nice one !
I saw `app/controllers` so thought it was weird

app/models/organisations_user.js
patrixr: weird that is shows it as an addition 🤔

app/controllers/users/charity_position.js
swatijadhav: maybe we can check if we can set some flags/settings in that, so that it will look into those things and ignore others.
https://prettier.io/docs/en/options.html

app/mixins/organisation.js
```diff
@@ -0,0 +1,43 @@
...
+import AsyncMixin from "./async";
+
+export default Ember.Mixin.create(AsyncMixin, {
+  organisationService: Ember.inject.service(),
+  /**
+   * Open's the organisation overlay.
+   * Returns the selected organisation record
+   * @returns {Object}
+   */
+  async organisationLookup() {
```
patrixr, line 11: solid name 👍

app/mixins/organisation.js
```diff
@@ -0,0 +1,43 @@
...
+      }
+
+      if (Object.keys(options).length) {
+        if (id) {
+          return this.replaceRoute(to, id, options);
+        }
+        return this.replaceRoute(to, options);
+      }
+    }
+  }
```
patrixr, line 42: The `organisation` variable is not used when `to` is present.

Meaning the user would see an overlay, but it would not navigate to which ever organisation they selected.

If an `id` is provided,  there is no need to call the lookup method at all

patrixr (approved):
Added one last comment, looks good otherwise

shreyas098 (approved):

swatijadhav (approved):

PR #871, opened 2020-09-11
GCW-3283 User Role Management Screens

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3283

### What does this PR do?

Screenshot 2020-09-11 at 4 34 08 PM
Screenshot 2020-09-11 at 4 35 44 PM
Screenshot 2020-09-11 at 4 36 07 PM


app/controllers/users/stock_roles.js
```diff
@@ -0,0 +1,189 @@
...
+
+      if (stockRoleAccess === "noAccess") {
+        this.get("userService").deleteStockRoles(this.get("user"));
+      } else {
+        if (this.get("canUpdateRole")) {
+          if (stockRoleAccess === "accessTill") {
+            roleExpiryDate = this.get("roleExpiryDate");
+          }
+
+          this.updateUserRole(
```
abulsayyad123, line 156: I think here you should send array of `permission` to the service and there you can iterate and update with other helper method.

app/services/user-service.js
```diff
@@ -0,0 +1,158 @@
...
+      .get("activeUserRoles")
+      .filter(
+        row => _.includes(roles, row.get("role.name")) && !!row.get("expiresAt")
+      )
+      .map(row => row.get("expiresAt"));
+
+    return _.max(expiryDates);
+  },
+
+  getRoleAccessText(user, app) {
```
bharat619, line 45: could this be
```javascript
  getRoleAccessText(user, userType){
    .....
  }
```
where userType could be like a stock or admin user?
Using an app name could confuse

app/services/user-service.js
```diff
@@ -0,0 +1,158 @@
...
+    }
+  },
+
+  deleteStockRoles(user) {
+    const userId = user.get("id");
+
+    if (this.canUpdateRole(userId)) {
+      const userRoleIds = user.get("roles").map(role => role.id);
+      const stockAppRoleIds = [
+        this.getRoleId("Stock administrator"),
```
bharat619, line 145: these strings have been used many times,
can move to any constant

app/components/calendar-input.js
```diff
@@ -7,6 +7,7 @@ const MIN_DATE = moment()
 const DEFAULT_PICKADATE_CONFIG = {
   selectMonths: true,
   selectYears: true,
+  format: "dd/mmm/yyyy",
```
patrixr, line 10: I notice there's `format` and `formatSubmit` 

I feel there should only be one, which is set by whoever uses the component 🤔

app/components/calendar-input.js
```diff
@@ -86,13 +87,22 @@ export default Ember.TextField.extend({
...
     }
+
+    const formatSubmit = this.get("formatSubmit");
+    if (formatSubmit) {
+      this.set("datePickerConfig", {
+        ...DEFAULT_PICKADATE_CONFIG,
+        format: formatSubmit
+      });
+    }
   },
```
patrixr, line 106: There's no need to `set(...)` 3 times : ) 

```suggestion
  initializeConfig() {
    let config = { ...DEFAULT_PICKADATE_CONFIG }

    const enablePastDate = this.get("enablePastDate");
    if (enablePastDate) {
      config.min = MIN_DATE;
    }

    const formatSubmit = this.get("formatSubmit");
    if (formatSubmit) {
      config.format = formatSubmit;
    }

    this.set("datePickerConfig", config);
  },
```

app/controllers/users/admin_roles.js
```diff
@@ -0,0 +1,154 @@
...
+  roleError: Ember.computed("noAdminAppRole", "adminRoleAccess", function() {
+    return (
+      this.get("noAdminAppRole") && this.get("adminRoleAccess") !== "noAccess"
+    );
+  }),
+
+  adminRoleAccess: Ember.computed("user.roles.[]", {
+    get() {
+      if (this.get("noAdminAppRole")) {
+        return "noAccess";
```
patrixr, line 43: rather than using strings, I'd recommend creating an enum

e.g

```javascript
const ACCESS_TYPES = {
     NO_ACCESS:               'noAccess',
     LIMITED_ACCESS:       'accessTill',
     UNLIMITED_ACCESS: 'accessForever'
}
```

app/controllers/users/stock_roles.js
patrixr: Hardcoding the role names definitely makes my spidey sense tingle. This should be derived from the list of roles you fetch from the API. 

If I add a new role on the backend, we should not have to update the UI at all to support it

app/services/user-service.js
```diff
@@ -0,0 +1,158 @@
...
+  session: Ember.inject.service(),
+  printerService: Ember.inject.service(),
+
+  adminAppRoles: ["Reviewer", "Supervisor"],
+  stockAppRoles: [
+    "Stock administrator",
+    "Stock fulfilment",
+    "Order administrator",
+    "Order fulfilment"
+  ],
```
patrixr, line 16: In theory, if this mapping was derived from the API. You could write the whole thing without ever hardcoding any role names

app/constants/roles.js
```diff
@@ -0,0 +1,13 @@
+export const ROLES = {
```
abulsayyad123, line 1: 👌

shreyas098 (approved):

bharat619 (approved):

PR #880, opened 2020-09-21
GCW-3282 disable/enable user account

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3282

### What does this PR do?
Enable/ Disable User


app/controllers/users/details.js
```diff
@@ -47,7 +53,42 @@ export default Ember.Controller.extend(OrganisationMixin, {
     }
   ),
 
+  disableUserAccount(disabled) {
```
abulsayyad123, line 56: If it disabling user, there is no need to pass false `params` . You can directly set `false` or If you are thinking for of using this function for both and enabling user then "ToggleUserAccount" make will make more sense i think.

app/controllers/users/details.js
```diff
@@ -47,7 +53,42 @@ export default Ember.Controller.extend(OrganisationMixin, {
...
+      this.runTask(async () => {
+        user.set("disabled", disabled);
+        await user.save();
+      }, ERROR_STRATEGIES.MODAL);
+    }
+  },
+
   actions: {
+    disableUser() {
+      this.disableUserAccount(true);
```
abulsayyad123, line 69: Named params makes more sense.. passing `true` or `false` is not much clear.

app/templates/users/details.hbs
```diff
@@ -19,6 +19,13 @@
     {{t "users.details.last_on_goodcity"}} {{time-ago timeValue=model.lastDisconnected smallPrefix=false}}
   </div>
 
+  {{#if (is-and canDisableUsers isDisabledUser)}}
```
abulsayyad123, line 22: if we can create computed property with meaningful name instead of using `and` condtion here would be much clearer 🤔  ?
Thoughts ?

app/styles/templates/_user_details.scss
```diff
@@ -268,3 +294,31 @@
   }
 }
 // END
+
+.disable-user-popup {
```
abulsayyad123, line 298: 👌

app/controllers/users/details.js
```diff
@@ -47,7 +53,42 @@ export default Ember.Controller.extend(OrganisationMixin, {
     }
   ),
 
+  disableUserAccount(disabled) {
+    if (this.get("canDisableUsers")) {
```
bharat619, line 57: would it be better to have any error message displayed, if user dont have this permission ?

app/controllers/users/details.js
bharat619: I think `disableUserAccount` sounds more like an action, rather than a property. I think it would be good to move the whole thing into `disableUser()` action written below

app/controllers/users/details.js
swatijadhav: The link to disable/enable account are only accessible to users who have permission to do it, otherwise, it will be always a hidden link.

app/controllers/users/details.js
```diff
@@ -47,7 +57,42 @@ export default Ember.Controller.extend(OrganisationMixin, {
     }
   ),
 
+  toggleUserAccount(options) {
+    if (this.get("canDisableUsers")) {
+      let user = this.get("user");
+
+      this.runTask(async () => {
```
bharat619, line 64: runTask is async, u might want to await it

bharat619 (approved):

abulsayyad123 (approved):

shreyas098 (approved):

PR #890, opened 2020-10-01
GCW-3236- Create creation form for international orders

shreyas098:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3236

### What does this PR do?

FEATURE: Overseas/International Order Creation.

### Screenshots

Screen Shot 2020-10-13 at 1 42 26 PM


app/controllers/order/international_order.js
```diff
@@ -0,0 +1,141 @@
...
+import SearchOptionMixin from "stock/mixins/search_option";
+import AsyncMixin, { ERROR_STRATEGIES } from "stock/mixins/async";
+import { regex } from "stock/constants/regex";
+
+export default Ember.Controller.extend(SearchOptionMixin, AsyncMixin, {
+  i18n: Ember.inject.service(),
+  showError: false,
+  orderService: Ember.inject.service(),
+
+  shipmentorCarryoutCode: Ember.computed({
```
abulsayyad123, line 11: Normal computed property will work no need to use `get` `set`.

app/controllers/order/international_order.js
```diff
@@ -0,0 +1,141 @@
...
+    get() {
+      return this.get("shipmentorCarryoutCode");
+    },
+    set(_, value) {
+      return value;
+    }
+  }),
+
+  isInvalidDescription: Ember.computed("orderDescription", function() {
```
abulsayyad123, line 20: use `computed.not`.

app/controllers/order/international_order.js
```diff
@@ -0,0 +1,141 @@
...
+
+  isInValidCountry: Ember.computed("country", function() {
+    return !this.get("country");
+  }),
+
+  isInvalidShipmentDate: Ember.computed("shipmentDate", function() {
+    return !this.get("shipmentDate");
+  }),
+
+  getCurrentUser: Ember.computed(function() {
```
abulsayyad123, line 40: I think there already exist `getCurrentUser` in service, if not create one and you directly access service method in `.hbs`.

app/controllers/order/international_order.js
```diff
@@ -0,0 +1,141 @@
...
+    var currentUser = store.peekAll("user_profile").get("firstObject") || null;
+    return currentUser;
+  }).volatile(),
+
+  shipmentTypes: Ember.computed(function() {
+    let translation = this.get("i18n");
+    let shipment = translation.t("order_transports.shipment");
+    let carryout = translation.t("order_transports.carry_out");
+
+    return [
```
abulsayyad123, line 51: move this to constant file.

app/controllers/order/international_order.js
```diff
@@ -0,0 +1,141 @@
...
+      ) {
+        this.set("showError", true);
+        return;
+      }
+
+      return this.runTask(async () => {
+        let data = await this.get("orderService").createShipmentorCarryoutOrder(
+          this.orderParams()
+        );
+        this.store.pushPayload(data);
```
abulsayyad123, line 124: handle this at service.

app/routes/order/international_order.js
```diff
@@ -0,0 +1,19 @@
+import Ember from "ember";
+import AuthorizeRoute from "./../authorize";
+
+export default AuthorizeRoute.extend({
+  orderService: Ember.inject.service(),
+
+  async model() {
+    return {
```
abulsayyad123, line 8: no need to use `async-await` in model. Model will automatically take care of it.

app/controllers/app_menu_list.js
```diff
@@ -51,8 +51,11 @@ export default Ember.Controller.extend(storageType, OrganisationMixin, {
         queryParams: { codeId: code.get("id"), storageType: storageTypeName }
       });
     },
+    createInternationalOrder() {
```
bharat619, line 54: doesnt sound right 🤔 

The method name is createInternationalOrder,
but its just changing the route

app/controllers/order/international_order.js
bharat619: yup.
can make use of app/service/session.js

app/controllers/order/international_order.js
```diff
@@ -0,0 +1,141 @@
...
+  },
+
+  orderParams() {
+    let validCode = this.formatOrderCode(
+      this.get("selectedType.id"),
+      this.get("shipmentorCarryoutCode")
+    );
+
+    let params = {
+      state_event: "submit",
```
bharat619, line 68: Its better to have strings like "submit", "Shipment", "CarryOut" etc to move into a constant file.
For example: if "submit" needs to be changed to "Submit", then it will be present in only one file, rather than changing in multiple file

app/controllers/order/international_order.js
```diff
@@ -0,0 +1,141 @@
...
+
+    typeSelection(value) {
+      return this.runTask(async () => {
+        this.set("selectedType", value);
+        let data = await this.get("orderService").fetchShipmentorCarryoutCode(
+          value.id
+        );
+        this.set("shipmentorCarryoutCode", data);
+      }, ERROR_STRATEGIES.MODAL);
+    },
```
bharat619, line 107:
```suggestion
    async typeSelection(value) {
      return await this.runTask(async () => {
        this.set("selectedType", value);
        let data = await this.get("orderService").fetchShipmentorCarryoutCode(
          value.id
        );
        this.set("shipmentorCarryoutCode", data);
      }, ERROR_STRATEGIES.MODAL);
    },
```

app/controllers/order/international_order.js
```diff
@@ -0,0 +1,141 @@
...
+        let data = await this.get("orderService").createShipmentorCarryoutOrder(
+          this.orderParams()
+        );
+        this.store.pushPayload(data);
+        this.clearFormData();
+        this.transitionToRoute("orders");
+      }, ERROR_STRATEGIES.MODAL);
+    },
+
+    onSearch(field, searchText) {
```
bharat619, line 130: i think u can remove this method completely and directly make use of `onSearchCountry` and try if doing something like below works

```
{{searchable-select on-search=onSearchCountry  }}
```

app/services/order-service.js
```diff
@@ -7,6 +7,18 @@ export default ApiBaseService.extend({
     return this.GET(`/users/${userId}/orders_count`);
   },
 
+  fetchShipmentorCarryoutCode(detail_type) {
+    return this.GET("/fetch_shipment_or_carryout_code", {
```
bharat619, line 11: you can just return the promise
`return this.GET("/fetch_shipment_or_carryout_code", { detail_type: detail_type })`
or
you can return the resolved data.
no need to return both

bharat619 (changes requested):

app/routes/order/international_order.js
shreyas098: I am getting a promise , so i am using async-await

abulsayyad123: `model` resolves promises automatically no need to explicitly resolve it :)

app/routes/order/international_order.js
```diff
@@ -0,0 +1,19 @@
...
+      shipmentSubsequentCode: await this.get(
```
bharat619, line 9: I guess you can completely remove the model here
Reason is that ur model method is actually not referring any model.
And you can move this to `setupController` completely

Remove this and make setupController as
```javascript
async setupController(controller, model) {
     const code  = await this.get("orderService").fetchShipmentorCarryoutCode("Shipment");
   controller.set("shipmentorCarryoutCode", code);
 }
```

app/routes/order/international_order.js
```diff
@@ -0,0 +1,19 @@
...
+        "orderService"
+      ).fetchShipmentorCarryoutCode("Shipment")
```
bharat619, line 11: "Shipment" and "Carrout" are also better placed in a constant, since the string is referred multiple places

bharat619 (commented):
Nice job. Some minor tweaks suggested

app/routes/order/international_order.js
shreyas098: okk... i missed that point

bharat619 (approved):

app/controllers/orders/contact_summary.js
```diff
@@ -1,15 +1,122 @@
...
+      codeRegEx.test(this.get("codeWithoutPrefix"))
+    );
+  }),
+
+  selectedCountry: Ember.computed("order", function() {
+    return {
+      id: this.get("order.countryId"),
+      nameEn: this.get("order.countryName")
+    };
+  }).volatile(),
```
abulsayyad123, line 30: why volatile ?

PR #891, opened 2020-10-05
GCW-3311- sorting & filtering of goods on an order view

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3311

### What does this PR do?

Add UI for sorting and filtering of Goods on Order details.


app/components/pluggable-drop-down.js
```diff
@@ -0,0 +1,16 @@
...
+  displayDropDownItems: false,
+
+  actions: {
+    toggleDisplayItems() {
+      this.toggleProperty("displayDropDownItems");
+    },
+
+    selectOption(item) {
+      this.get("setSelectedValue")(item);
+      this.send("toggleDisplayItems");
```
patrixr, line 13:
```suggestion
      this.set("displayDropDownItems", false);
```

in case the dropdown is closed when the action is called (you never know)

app/constants/order-sorting-options.js
```diff
@@ -0,0 +1,44 @@
...
+    column_name: "Inventory ID",
+    column_alias: "packages.inventory_number",
+    sort: "desc",
+    is_desc: true
+  },
+  {
+    column_name: "Added Date",
+    column_alias: "orders_packages.created_at",
+    sort: "asc",
+    is_desc: false
```
patrixr, line 18: `is_desc` is redundant with the sort value, you can probably remove it

app/constants/order-sorting-options.js
```diff
@@ -0,0 +1,44 @@
+export const ORDER_SORTING_OPTIONS = [
```
patrixr, line 1: maybe rename the constant (and file) to `BOX_CONTENT_SORTING_OPTIONS`
Because you're not actually sorting orders here

app/controllers/orders/active_items.js
```diff
@@ -18,12 +27,51 @@ export default detail.extend({
     "sortProperties"
   ),
 
+  setDefaultSortingColumn: Ember.computed("dropDownItems", {
+    get(key) {
+      return ORDER_SORTING_OPTIONS[0];
+    },
+    set(key, value) {
+      return value;
+    }
```
patrixr, line 36: that's a setter that doesn't actually set anything, and the property's name `setXXX` makes it feel like a function.

it seems dependant on `dropDownItems` without every using that property which is weird.
Not sure what the intended behaviour is, but if you want a property that gets modified later, you can literally just do

```
sortingColumn: ORDER_SORTING_OPTIONS[0], // initial value

// and later

set('sortingColumn', column)
```

app/constants/order-sorting-options.js
```diff
@@ -0,0 +1,44 @@
...
+  {
+    column_name: "Inventory ID",
+    column_alias: "packages.inventory_number",
```
patrixr, line 4: `column_key` makes more sense than "alias"

app/controllers/orders/active_items.js
```diff
@@ -18,12 +27,51 @@ export default detail.extend({
...
+  stateList: Ember.observer("states.@each.enabled", function() {
+    this.reloadResults();
+    this.set("filteredStates", this.getFilteredStates());
+  }),
+
+  getSortQuery() {
+    const sortOptions = this.get("setDefaultSortingColumn");
+    return {
+      sort_column: sortOptions["column_alias"],
+      is_desc: sortOptions["is_desc"]
```
patrixr, line 62:
```suggestion
      is_desc: /desc/.test(sortOptions["sort"])
```

you can also make it a computed property

```
sortQuery: Ember.computed('sortOptions', function () {
  const { column_alias, sort } = this.get("sortOptions");
  return {
    sort_column: column_alias,
    is_desc: /desc/i.test(sort)
  }
})
```

app/controllers/orders/active_items.js
```diff
@@ -34,6 +82,11 @@ export default detail.extend({
         });
     },
 
+    applySortOn(options) {
+      this.set("setDefaultSortingColumn", options);
```
patrixr, line 86: notice the duplication of "set" ... that's why the name of this property is weird

app/templates/orders/active_items.hbs
```diff
@@ -24,11 +24,96 @@
...
+      id="searchText"
+      class="search-orders"
+      value=searchText }}
+      {{#if searchText.length}}
+        <i {{action "clearSearch"}} class="cancel-search">{{fa-icon 'times-circle'}}</i>
+      {{/if}}
+    </div>
+    <div class="small-4 columns">{{t "order_details.orders_packages.sort_by"}}:
+      {{#pluggable-drop-down setSelectedValue=(action "applySortOn") as |dropDown|}}
+        {{#dropDown.selectedDropDown}}
```
patrixr, line 41: handlebars slots, nice 👍  !

app/templates/components/pluggable-drop-down.hbs
```diff
@@ -0,0 +1,9 @@
+<div {{action "toggleDisplayItems"}} >
```
patrixr, line 1: I'd argue that `composable-drop-down` is a better name, but I'm not fussy.
very generic and reusable, well done ! 👍 

since you're using `dropDownBody` as a slot name, I'd recommend `dropDownHeader` instead of `selectedDropDown`. 

Body and Header are a common combination, so people would expect it, and `selectedDropDown` doesn't sound like a slot name, it sounds like a property of the component

app/controllers/orders/active_items.js
abulsayyad123: Makes sense and its much cleaner.. Thanks.

app/controllers/orders/active_items.js
```diff
@@ -18,12 +27,51 @@ export default detail.extend({
...
+    };
+  },
+
+  getFilteredStates() {
+    return _.filter(this.get("states"), ["enabled", true]).map(
+      ({ state }) => state
+    );
+  },
+
+  stateList: Ember.observer("states.@each.enabled", function() {
```
patrixr, line 53: you name the property `stateList` but it's not actually a list, it's an observer.

Maybe something like `stateChangeListener` is more appropriate

app/controllers/orders/active_items.js
```diff
@@ -18,12 +27,51 @@ export default detail.extend({
...
+    return {
+      state
+    };
+  },
+
+  getFilteredStates() {
+    return _.filter(this.get("states"), ["enabled", true]).map(
+      ({ state }) => state
+    );
+  },
```
patrixr, line 51: I think you can just make a `filteredStates` computed property, and you wouldn't need an observer to listen and set it

patrixr (changes requested):
Few comments here and there, but looks good overall ! Good one !

app/controllers/orders/active_items.js
```diff
@@ -1,29 +1,74 @@
...
+
+  // ----- Computed Properties -----
   sortedGcRequests: Ember.computed.sort(
     "model.goodcityRequests",
     "sortProperties"
   ),
 
+  filteredStates: Ember.computed("states.@each.enabled", function() {
+    this.reloadResults();
+    return this.getFilteredStates();
```
patrixr, line 36:
```suggestion
    return _.filter(this.get("states"), ["enabled", true]).map(
      ({ state }) => state
    );
```

You can put the body straight in, and remove the `getFilteredStates` method

patrixr (approved):

shreyas098 (approved):

PR #898, opened 2020-10-07
GCW-3314 designate confirmation screen

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3314

### What does this PR do?

designate confirmation screen

Screenshot 2020-10-07 at 5 38 32 PM
Screenshot 2020-10-07 at 6 00 54 PM

Screenshot 2020-10-07 at 5 40 00 PM


app/components/goodcity/designate-form.js
```diff
@@ -15,6 +21,14 @@ export default Ember.Component.extend({
...
   }),
 
+  onOrderChange: Ember.observer("order", function() {
+    this.set("quantity", this.get("maxQuantity"));
+  }),
+
+  onItemChange: Ember.observer("pkg", function() {
+    this.set("quantity", this.get("maxQuantity"));
+  }),
+
```
patrixr, line 31:
```suggestion
  onRecordsChange: Ember.observer("order", "pkg", function() {
    this.set("quantity", this.get("maxQuantity"));
  }),
```

app/components/goodcity/designate-form.js
```diff
@@ -52,5 +66,33 @@ export default Ember.Component.extend({
...
+  ),
+
+  actions: {
+    displayItemAvailabilityOverlay() {
+      this.set("showItemAvailability", true);
+    },
+
+    closeItemAvailabilityOverlay() {
+      this.set("showItemAvailability", false);
+    },
```
bharat619, line 78: can have a single function

```javascript
setItemShowAvailibility(val) {
 this.set('showItemAvailability', val)
}
```

app/templates/components/goodcity/designate-link.hbs
```diff
@@ -1,17 +1,5 @@
...
-    {{goodcity/designate-form
-      quantity=designationQty
-      maxQuantity=designatableQuantity
-      editableQty=editableQty
-      confirmAction=(action "completeDesignation")
-      cancelAction=(action "cancelDesignation")
-      order=designationTargetOrder
-      pkg=designationTargetPackage
-    }}
-  {{/if}}
```
patrixr, line 16: This component had the whole designation flow, not it only has half. Feels like a downgrade.
With this change, it'll be very hard for someone to use it without knowing the intricate details of how it works.

If this component is not appropriate for use in the page you're building it's probably best not to use it and use a normal button with your custom actions. This feels too damaging

app/components/goodcity/designate-form.js
```diff
@@ -52,5 +66,33 @@ export default Ember.Component.extend({
...
+      this.set("showItemAvailability", false);
+      this.set("readyToDesignate", false);
+      this.router.transitionTo("items.detail.publishing", this.get("pkg.id"));
+    },
+
+    modifyBoxPalletAllocation() {
+      this.set("showItemAvailability", false);
+      this.set("readyToDesignate", false);
+      this.router.transitionTo("items.detail.location", this.get("pkg.id"));
+    }
```
bharat619, line 96: can it be done something like this? since only the routing changes and rest all are same

```suggestion
    action(name){
        this.set("showItemAvailability", false);
        this.set("readyToDesignate", false);
        switch(name){
           case 'gain':
              this.router.transitionTo("items.detail.publishing", this.get("pkg.id"));
              break;
           case 'modifyDesignation':
              this.router.transitionTo("items.detail.publishing", this.get("pkg.id"));
              break;
           case 'modifyBoxPalletAllocation':
             this.router.transitionTo("items.detail.location", this.get("pkg.id"));
             break;
        }
    
   
```

app/templates/components/goodcity/designate-link.hbs
patrixr: Oh nevermind, I just saw the global component. Ballsy move, but high reward if it works out well

app/services/designation-service.js
```diff
@@ -171,6 +176,108 @@ export default ApiBaseService.extend(NavigationAwareness, {
...
+    if (pkg.get("availableQuantity") > 0 || options.isEditOperation) {
+      this.triggerDesignateAction(pkg, order);
+    } else {
+      this.set("designationTargetPackage", pkg);
+      this.set("displayDesignateForm", true);
+      this.set("showItemAvailability", true);
+    }
+  },
+
+  triggerDesignateAction(pkg, order) {
```
patrixr, line 204: can you maybe prefix any private method with and underscore (`_something`).
To make sure everyone knows what the entry points are to those global functionalities (and avoid people calling helper methods directly)

app/services/designation-service.js
```diff
@@ -171,6 +176,108 @@ export default ApiBaseService.extend(NavigationAwareness, {
...
   },
 
+  async resolveOrder(order) {
+    if (order) {
+      return order;
+    }
+    return this.userPickOrder({
+      state: ACTIVE_ORDER_STATES.join(",")
+    });
+  },
```
patrixr, line 186:
```suggestion
  async resolveOrder(order) {
    return order || this.userPickOrder({
      state: ACTIVE_ORDER_STATES.join(",")
    });
  },
```

unimportant, just an alternative

app/services/designation-service.js
```diff
@@ -25,14 +27,16 @@ function ID(modelOrId) {
...
+export default ApiBaseService.extend(NavigationAwareness, AsyncMixin, {
   store: Ember.inject.service(),
 
   init() {
     this._super(...arguments);
     this.set("openOrderSearch", false);
   },
 
+  editableQty: Ember.computed.alias("settings.allowPartialOperations"),
+
```
patrixr, line 39: I'm wondering if those are needed now that we've retired stockit, not a big deal to have it though

app/services/designation-service.js
```diff
@@ -171,6 +176,108 @@ export default ApiBaseService.extend(NavigationAwareness, {
...
+      this.triggerDesignateAction(pkg, order);
+    } else {
+      this.set("designationTargetPackage", pkg);
+      this.set("displayDesignateForm", true);
+      this.set("showItemAvailability", true);
+    }
+  },
+
+  triggerDesignateAction(pkg, order) {
+    this.resolveOrder(order).then(target => {
```
bharat619, line 205: Assuming that `resolveOrder` just brings up an overlay, I think it will be better to do 
```
const target = await this.resolveOrder(order)
if(!target) return
// rest of the logic

```

patrixr (approved):
Looks good ! A global designation UI is a big change, but given the state of the app in general I think there's more value in taking some risks like this rather than playing it safe :)

nice one !

app/services/designation-service.js
```diff
@@ -171,6 +176,108 @@ export default ApiBaseService.extend(NavigationAwareness, {
...
+    this.set("designatableQuantity", 0);
+    this.set("designationTargetPackage", null);
+    this.set("designationTargetOrder", null);
+    this.set("designationQty", 0);
+  },
+
+  completeDesignation() {
+    if (!this.get("readyToDesignate")) return;
+
+    this.runTask(() => {
```
bharat619, line 265: runTask is async function

```suggestion
    await this.runTask(() => {
```

app/services/designation-service.js
```diff
@@ -171,6 +176,108 @@ export default ApiBaseService.extend(NavigationAwareness, {
...
+      const quantity = this.get("designationQty");
+      const shippingNumber = this.get("shippingNumber");
+
+      return this.designate(pkg, {
+        order,
+        quantity,
+        shippingNumber
+      });
+    }, ERROR_STRATEGIES.MODAL).finally(() => {
+      this.cancelDesignation();
```
bharat619, line 277: since this is in finally block,
will this not cancel the designation even if it successfully does the `completeDesignation` 🤔

bharat619: oh just saw the `cancelDesignation` method, may be rename it to a different method like `resetOrderDesignation` or `initializeOrderDesignation`

bharat619 (commented):
Pointed out some minor changes, rest all looks fine

bharat619 (approved):

shreyas098 (approved):

app/components/goodcity/designate-form.js
swatijadhav: Adding multiple properties in the observer does not support it. Verified above suggested change on local.

PR #901, opened 2020-10-08
GCW-3347 fixed item action icons on history page

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3347

### What does this PR do?

fixed item action icons on history page

### DIFF
--- a/app/constants/action-icons.js
+++ b/app/constants/action-icons.js
@@ -1,0 +1,34 @@
+export const ACTIONS_ICONS = [
+  {
+    name: "Process",
+    icon: "random"
+  },
+  {
+    name: "Recycle",
+    icon: "recycle"
+  },
+  {
+    name: "Trash",
+    icon: "dumpster"
+  },
+  {
+    name: "Gain",
+    icon: "folder-plus"
+  },
+  {
+    name: "Loss",
+    icon: "folder-minus"
+  },
+  {
+    name: "Edited",
+    icon: "pencil-alt"
+  },
+  {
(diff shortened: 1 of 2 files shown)


shreyas098 (approved):

app/constants/action-icons.js
```diff
@@ -0,0 +1,34 @@
+export const ACTIONS_ICONS = [
```
patrixr, line 1: This seems to be a duplicate of `constants/item-actions`

Also, `Edited`, `Inventory` and `Pack` shouldn't be in the bottom menu.

`Edited` isn't a valid inventory action at all

patrixr: Can you point out what the actual change is ?

patrixr: Also note that the title of the PR is "fix icons on history page", but it's modifying a helper that can be used _everywhere_

swatijadhav: `constants/item-actions` this is specifically used in the actions dropdown next to designate button on item-published tab. And those are specific to item gain-loss.

On history tab, the actions like `pack`, `Edited`(groups item attribute updates), `inventory` are listed which are missing in  `constants/item-actions` and hence added different files only specific for icons.

patrixr: Aaah, gotcha ! I thought these were for the bottom menu. Looks good then :)

patrixr (approved):

PR #904, opened 2020-10-09
GCW-3348 Box Pallet Infinite Scroll

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3348

### What does this PR do?
- Adds infinite scroll for items in box or pallet
- Improve on existing actions by adding callbacks to handle addition of items to box-pallets / removal of items from box-pallets in the infinite-scroll section

### DIFF
--- a/app/components/contained-package.js
+++ b/app/components/contained-package.js
@@ -15,6 +15,7 @@ export default Ember.Component.extend(ItemActionMixin, {
 
   fetchAddedQuantity: Ember.computed("pkg", function() {
     let pkgId = this.get("pkg.id");
+
     let entityId = this.get("entity.id");
     let promise = this.get("packageService").fetchAddedQuantity(
       entityId,
--- a/app/components/contained-package.js
+++ b/app/components/contained-package.js
@@ -1,8 +1,6 @@
 import Ember from "ember";
 import ItemActionMixin from "stock/mixins/item_actions";
 
-const ObjectPromiseProxy = Ember.ObjectProxy.extend(Ember.PromiseProxyMixin);
-
 export default Ember.Component.extend(ItemActionMixin, {
   packageService: Ember.inject.service(),
   store: Ember.inject.service(),
@@ -11,22 +9,5 @@ export default Ember.Component.extend(ItemActionMixin, {
     return this.get("store").peekRecord("item", this.get("pkg.id"));
   }),
 
-  disableRemove: Ember.computed.alias("entity.isDispatched"),
-
-  fetchAddedQuantity: Ember.computed("pkg", function() {
-    let pkgId = this.get("pkg.id");
-
-    let entityId = this.get("entity.id");
-    let promise = this.get("packageService").fetchAddedQuantity(
-      entityId,
-      pkgId
-    );
-    return ObjectPromiseProxy.create({
-      promise
-    });
-  }),
-
-  addedQuantityCount: Ember.computed
-    .reads("fetchAddedQuantity.added_quantity")
-    .readOnly()
+  disableRemove: Ember.computed.alias("entity.isDispatched")
 });
--- a/app/components/goodcity/add-item-overlay.js
+++ b/app/components/goodcity/add-item-overlay.js
@@ -38,16 +38,11 @@ export default Ember.Component.extend(AsyncMixin, {
   ),
 
   async resolveAddItemPromises() {
-    try {
-      await this.runTask(async () => {
-        const promises = this.addItemPromises();
-        await Ember.RSVP.all(promises);
-        this.sendAction("onConfirm");
-        this.set("open", false);
-      }, ERROR_STRATEGIES.RAISE);
-    } catch (err) {
-      throw error;
-    }
+    await this.runTask(async () => {
+      const promises = this.addItemPromises();
+      await Ember.RSVP.all(promises);
+      this.sendAction("onConfirm");
+    }, ERROR_STRATEGIES.RAISE);
   },
 
   hasInvalidAddedQuantity() {
@@ -82,19 +77,21 @@ export default Ember.Component.extend(AsyncMixin, {
   },
 
   actions: {
-    async moveItemToBox(pkg, cb) {
-      if (pkg) {
-        if (this.hasInvalidAddedQuantity()) {
-          this.get("messageBox").alert(
-            this.get("i18n").t("box_pallet.invalid_quantity")
-          );
-        } else {
-          await this.resolveAddItemPromises();
-          if (cb) {
-            cb(pkg);
-          }
-        }
+    async moveItemToBox(pkg, cb = _.noop) {
+      if (!pkg) {
+        return;
       }
+
+      if (this.hasInvalidAddedQuantity()) {
+        this.get("messageBox").alert(
+          this.get("i18n").t("box_pallet.invalid_quantity")
+        );
+      }
+
+      await this.resolveAddItemPromises();
+      cb(pkg, this.get("totalNumberTomove"));
+      this.set("open", false);
+      return pkg;
     },
 
     cancelMove() {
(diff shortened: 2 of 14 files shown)


app/components/contained-package.js
```diff
@@ -1,8 +1,6 @@
 import Ember from "ember";
```
patrixr, line 1: I like seeing a lot of red

app/components/goodcity/add-item-overlay.js
```diff
@@ -35,17 +37,17 @@ export default Ember.Component.extend({
...
+      await this.runTask(async () => {
+        const promises = this.addItemPromises();
+        await Ember.RSVP.all(promises);
         this.sendAction("onConfirm");
         this.set("open", false);
-      })
-      .finally(() => loadingView.destroy());
+      }, ERROR_STRATEGIES.RAISE);
+    } catch (err) {
+      throw error;
```
patrixr, line 49: no point in catching an error just to throw it again.
Plus, the error is named `err` not `error`

patrixr: just remove the try catch and let it propagate if that's the intention

app/components/goodcity/add-item-overlay.js
```diff
@@ -80,15 +82,17 @@ export default Ember.Component.extend({
...
           this.get("messageBox").alert(
             this.get("i18n").t("box_pallet.invalid_quantity")
           );
         } else {
-          this.resolveAddItemPromises();
+          await this.resolveAddItemPromises();
+          if (cb) {
+            cb(pkg);
+          }
         }
```
patrixr, line 96:
```suggestion
    async moveItemToBox(pkg, cb = _.noop) {
      if (!pkg) return;
      if (this.hasInvalidAddedQuantity()) {
        return this.get("messageBox").alert(
          this.get("i18n").t("box_pallet.invalid_quantity")
        );
      }
     
     await this.resolveAddItemPromises();
     cb(pkg);
     return pkg; // return it anyways, that way you support both callbacks and promises
  }
```

Untested, but something like that would be an improvement. You can support both callbacks and promises. 
Tip: using `_.noop` as a default value removes the need for an `if {}` statement

Also `resolveAddItemPromises` is a bit confusing...not sure what an "addItemPromise" is 🤔

app/controllers/items/detail.js
```diff
@@ -573,15 +569,55 @@ export default GoodcityController.extend(
...
-            .fetchContainedPackages(this.get("item.id"))
-            .then(data => {
-              this.get("store").pushPayload(data);
-              this.set("associatedPackages", data.items);
-            })
+      fetchContainedPackages(page = 1) {
+        // Set the associatedPackages to []
+        // This is to flush out all the old records iff page is 1
+        if (page === 1) {
+          this.set("associatedPackages", []);
```
patrixr, line 576: maybe the full list that exists inside the infinite-scroll component could be made accessible to the controller for these kinds of operations. To avoid you having to work around it. Maybe check with @abulsayyad123

patrixr: you could also have your `fetchContainedPackages` function return wrapped records, that way you can add your own properties (locally) to mark them as deleted.

e.g

the function returns a list like this

```javascript
[
   {
       record: record,
       deleted: false
   }
   ...
]
```

After a successful deletion you can set `deleted` to `true`.

And in the handlebars file you can do

```
{{#unless row.deleted}}
   <div> ... </div>
{{/unless}}
```

That way you wouldn't need to store a "parallel" list to the component's

patrixr: Actually... I just checked the ember docs, and it seems records all have an `isDeleted` property.

So simply by having a handlebars `{{#unless}}` you can hide them from the UI without needing to keep track of a list

https://guides.emberjs.com/release/models/creating-updating-and-deleting-records/

app/controllers/items/detail.js
```diff
@@ -624,14 +665,36 @@ export default GoodcityController.extend(
...
             }
+            return data;
           });
       },
 
       openItemsSearch() {
         this.set("openPackageSearch", true);
       },
 
+      async updateAssociatedPackages(pkg) {
```
patrixr, line 684: I think the terminology `associated packages` is confusing. Should always be either `contained packages` or `container`

app/mixins/item_actions.js
```diff
@@ -113,35 +113,33 @@ export default Ember.Mixin.create(AsyncMixin, {
...
+    };
+
+    // 1. Make API request to remove item from the container
+    this.get("packageService")
+      .addRemoveItem(container.id, params)
+      .then(() => {
+        this.get("packageService");
+
+        // 2. Reload the model to sync all location and quantity data with API
+        item.reload();
```
patrixr, line 134: you probably need to await that before calling the callback. I'd recommend turning the whole method into an async method

app/templates/components/goodcity/item-search-overlay.hbs
```diff
@@ -26,7 +26,7 @@
             {{#if displayResults}}
               {{#infinite-list height="85vh" loadMore=(action "loadMoreItems") as |items| }}
                 {{#each items as |item|}}
-                  <div class="row" {{action 'selectItem' item}}>
+                  <div class="row" {{action 'selectItem' item cb}}>
```
patrixr, line 29: if your goal with the callback is to have multiple actions trigger one after the other, you can check out the `multi-actions` helper I created in stock

works like this

```handlebars
 <button {{action (multi-actions
    (action "doThis")
    (action "doThat")
    (action "andThenThis")
  )
 }}>
```

patrixr: note: actions need to return a promise for it work work well. So `async` is encouraged

patrixr (commented):
Looks good, although I did mark a few spots for potential improvement : )

app/templates/components/goodcity/item-search-overlay.hbs
patrixr: FYI: completely optional, mentioning it just in case

app/components/goodcity/add-item-overlay.js
bharat619: Yeah its a bit difficult to understand it. Would love to rewrite this file, but it just increases the scope. So I kept it as it is for now

app/controllers/items/detail.js
bharat619: I made it to return `isDeleted` 
The package-service method doesn't return an array of EmberObject, but the response from the API, so it wont have the isDeleted

I added ur first suggestion : )

app/templates/components/goodcity/item-search-overlay.hbs
bharat619: adding callback here was not necessary. I have updated it

app/components/goodcity/add-item-overlay.js
```diff
@@ -80,17 +77,21 @@ export default Ember.Component.extend({
...
+      }
+
+      if (this.hasInvalidAddedQuantity()) {
+        this.get("messageBox").alert(
+          this.get("i18n").t("box_pallet.invalid_quantity")
+        );
       }
+
+      await this.resolveAddItemPromises();
+      cb(pkg, this.get("totalNumberTomove"));
```
abulsayyad123, line 92: How callback is going to work in action ? Not sure how we are going to call it ?

bharat619: A template can have something like this
```
  {{some-component model=model value=val cb=(action 'handleChange'}}
```
In the template's controller there might be an action
```
handleChange(){
  // .. some code to handle change
}
```
So, in some-component.js it can be invoked as `cb()`, a callback so the controller can tell what to do after some action has been done.

The purpose of using a callback here is that overlay component acts independently and by using a callback system there is a way of controller doing something after overlay component has finished its action

patrixr (approved):

abulsayyad123 (approved):

PR #922, opened 2020-10-27
FIX=> willDestroy hook added for inline-components

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3360

### What does this PR do?
Added `willDestroyElement` hook for all inline edit components.

### DIFF
--- a/app/components/additional-fields-input.js
+++ b/app/components/additional-fields-input.js
@@ -44,5 +44,9 @@ export default Ember.TextField.extend({
 
   click() {
     this.addCssStyle();
+  },
+
+  willDestroyElement() {
+    this.destroy();
   }
 });
--- a/app/components/beneficiary-inline-input.js
+++ b/app/components/beneficiary-inline-input.js
@@ -82,5 +82,9 @@ export default Ember.TextField.extend({
   click() {
     this.addCssStyle();
     this.set("previousValue", this.get("value") || "");
+  },
+
+  willDestroyElement() {
+    this.destroy();
   }
 });
--- a/app/components/beneficiary-inline-number-input.js
+++ b/app/components/beneficiary-inline-number-input.js
@@ -1,6 +1,6 @@
-import Ember from 'ember';
-import AjaxPromise from 'stock/utils/ajax-promise';
-import BeneficiaryInlineInput from './beneficiary-inline-input';
+import Ember from "ember";
+import AjaxPromise from "stock/utils/ajax-promise";
+import BeneficiaryInlineInput from "./beneficiary-inline-input";
 const { getOwner } = Ember;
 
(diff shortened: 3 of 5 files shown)


app/components/additional-fields-input.js
```diff
@@ -44,5 +44,9 @@ export default Ember.TextField.extend({
 
   click() {
     this.addCssStyle();
+  },
+
+  willDestroyElement() {
+    this.destroy();
```
bharat619, line 50: Interesting 🤔 
Curious to know, Why should `this.destroy` be called explicitly? 

Isnt it the case that when component is being removed from DOM, the `willDestroyElement` is invoked?
So the point that its invoking `willDestroyElement` is it will be executing `this.destroy()` next.

In short, `willDestroyElement ` is a clean up method hook before `this.destroy` is invoked
Isn't that right?

abulsayyad123: In ember we have to explicitly destroy the component. When we were doing browser back button, the component was still in the DOM giving reference error, also it is better to destroy the component if you are leaving the page to avoid memory leaks.

patrixr: It does feel weird

patrixr: What was the original bug behind this fix ?

abulsayyad123: We are doing ajax request on focusOut in component and when were going out of this page with browser back at that time  ajax was in process to cancel it, i had to explicity destroy the element.

bharat619 (approved):

app/components/additional-fields-input.js
patrixr: I don't think this fix is right, 

but maybe you can try something like

```javascript
  focusOut() {
    if (this.get('isDestroying') || this.get('isDestroyed')) return;

    const config = {
      value: this.get("value"),
      name: this.get("name"),
      previousValue: this.get("previousValue")
    };
    this.get("onFocusOut")(config);
    this.set("inlineTextInput", false);
  },
```

ref: https://api.emberjs.com/ember/3.21/classes/Component/properties/isDestroying?anchor=isDestroying

Darshan-Chauhan: Sorry for dropping unannounced but component will get destroy once route is changed on clicking back link not before that so it won't prevent component from sending AJAX request. 
(Assuming the case is clicking browser back while still editing the field)

I think a good solution to this would be checking if event => target object has any DOM elements while focusing out/clicking on something else

```
focusOut(event) {

}
```

CC: @abulsayyad123

abulsayyad123:
Closing this PR. Since, issue has been fixed in other PRs.

PR #936, opened 2020-11-11
GCW-2626 Redesignation fixes

patrixr:
Fixes:

- Redesignate defaults to the same order, user can later change it
- Issue in computing max quantity

### DIFF
--- a/app/templates/components/goodcity/orders-package-block.hbs
+++ b/app/templates/components/goodcity/orders-package-block.hbs
@@ -57,4 +57,18 @@
         {{action-drawer isOpen=drawerOpened actionList=actionList small=small}}
     </div>
   {{/unless}}
+
+  {{#unless readOnly}}
+    {{goodcity/dispatch-form
+      open=readyToDispatch
+      orderPkg=dispatchOrdersPackage
+      location=dispatchLocation
+      quantity=dispatchQty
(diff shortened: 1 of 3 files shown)


app/templates/components/goodcity/orders-package-block.hbs
```diff
@@ -57,18 +57,4 @@
         {{action-drawer isOpen=drawerOpened actionList=actionList small=small}}
     </div>
   {{/unless}}
-
```
patrixr, line 60: @swatijadhav Can you confirm that removing this is okay ? Since we're using the new service entry point you created

swatijadhav: @patrixr this code seems to be for dispatch form. In the designation-confirmation changes PR https://github.com/crossroads/stock.goodcity/pull/898 we have already removed code changes for designation overlay here
https://github.com/crossroads/stock.goodcity/pull/898/files#diff-4365678f4a18a6be110bcf81c2f18a4d5b7a8aa3d1f6447d597e0c6b2d4cc911

Can you please have a look?

patrixr: ooooh thanks : D

swatijadhav (approved):

PR #939, opened 2020-11-12
Adding Package Guidance Feedback

bharat619:
Adding Package Guidance Feedback for https://jira.crossroads.org.hk/browse/GCW-3380 and https://jira.crossroads.org.hk/browse/GCW-3381


app/controllers/items/detail.js
```diff
@@ -434,8 +434,15 @@ export default GoodcityController.extend(
         this.set("selectedDescriptionLanguage", langauge);
       },
 
-      async updateAttribute(name, value, cb = _.noop) {
+      /**
+       * @param {String} name | Name of the field to update
+       * @param {any} value | Value of the field
+       * @param {boolean} isRequired | false by default
```
patrixr, line 440: FYI: JSDoc supports [optional params and default args](https://jsdoc.app/tags-param.html#optional-parameters-and-default-values) ;)

The brackets `[]` indicate that it's optional and the `=false` indicates the default value

```javascript
       * @param {boolean} [isRequired=false]
```

app/controllers/items/detail.js
```diff
@@ -434,8 +434,15 @@ export default GoodcityController.extend(
...
+       * @param {function} cb | Callback to invoke
+       */
+      async updateAttribute(name, value, isRequired = false, cb = _.noop) {
         const item = this.get("item");
+        if (!value && isRequired) return;
```
patrixr, line 445: A bit weird, this would prevent setting a value to `null` or `""`

app/controllers/items/detail.js
```diff
@@ -678,15 +685,24 @@ export default GoodcityController.extend(
...
         }
-        await this.saveItem(item);
-        this.send("setShowDescSuggestion", false);
+        await this.runTask(async () => {
+          await this.get("packageService").updatePackage(item, {
+            package: {
+              [param]: description
+            }
+          });
+        }, ERROR_STRATEGIES.MODAL);
```
patrixr, line 705: you can probably use `updateAttribute` here

app/controllers/items/detail.js
bharat619: It will just prevent if value is not present and the field is required
In case value needs to be null or "" then isRequired will be false by default, so it would not affect that

patrixr: I see, I would suggest `allowEmpty` instead of `isRequired` but no biggie

patrixr (approved):

PR #948, opened 2020-11-24
GCW-3371  User can remove specific quantity from box or pallet

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3371

### What does this PR do?
User can remove specific quantity from box or pallet


shreyas098 (approved):

app/components/goodcity/remove-item-from-container.js
```diff
@@ -0,0 +1,58 @@
...
+        this.set("maxRemovableQuantity", quantity);
+
+        this.set("location", selectedLocation);
+        this.set("openRemoveItemOverlay", true);
+      }
+    },
+
+    async performUnpack() {
+      await this.runTask(() => {
+        this.unpack(
```
patrixr, line 42:
```suggestion
        return this.unpack(
```

Otherwise the promise can't be caught :)

app/mixins/item_actions.js
```diff
@@ -212,32 +212,6 @@ export default Ember.Mixin.create(AsyncMixin, {
...
-
-    /**
-     * Unpack the requested quantity from a container (either box or pallet)
-     * and invoke any callback passed as argument.
-     * @param EmberObject container
-     * @param EmberObject item
-     * @param Integer quantity
-     * @param function callback
-     */
-    async unpack(container, item, quantity, callback) {
```
bharat619, line 225: I would suggest to keep this function
Reason: 
`unpack` always needs to be done to a location. Having the location logic inside a component/controller will be repetitive. And hence the location selection and unpacking logic moved to mixin, so that its easily pluggable to any component/controller.

If you need to return true/false, it would be better to return a `false` instead of doing just return
```javascript
const selectedLocation = await this.get(
        "locationService"
      ).userPickLocation();

      if (!selectedLocation) {
        return false;
      }
```

bharat619 (approved):

PR #950, opened 2020-11-24
GCW-3408 Disabled user account can have both email & phone removed

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3408

### What does this PR do?
- A disabled user account can have both email & phone removed
- DIsplay warning to update user on re-enable if email and mobile are blank


shreyas098 (approved):

app/controllers/users/contact_details.js
```diff
@@ -12,15 +12,40 @@ export default Ember.Controller.extend(
...
+    isValidEmail(email) {
+      return email.match(new RegExp(regex.EMAIL_REGEX));
+    },
 
-    invalidMobile: Ember.computed("mobileNumber", function() {
-      const hkMobileNumberRegEx = new RegExp(regex.HK_MOBILE_NUMBER_REGEX);
-      return this.get("mobileNumber").match(hkMobileNumberRegEx);
-    }),
+    isValidMobile(mobile) {
+      return mobile.match(new RegExp(regex.HK_MOBILE_NUMBER_REGEX));
```
bharat619, line 20: if this needs to return true/false,
then `return regex.HK_MOBILE_NUMBER_REGEX.test(mobile)` should work and can get rid of `Boolean(this.isValidMobile(this.get("mobileNumber")))` this line in the below method

app/controllers/users/contact_details.js
```diff
@@ -12,15 +12,40 @@ export default Ember.Controller.extend(
     user: Ember.computed.alias("model.user"),
     userService: Ember.inject.service(),
 
-    invalidEmail: Ember.computed("user.email", function() {
-      const emailRegEx = new RegExp(regex.EMAIL_REGEX);
-      return this.get("user.email").match(emailRegEx);
-    }),
+    isValidEmail(email) {
+      return email.match(new RegExp(regex.EMAIL_REGEX));
```
bharat619, line 16: if this needs to return true/false,
then return `regex.EMAIL_REGEX.test(email)` should work and can get rid of `Boolean(this.isValidMobile(this.get("mobileNumber")))` this line in the below method

bharat619 (approved):

PR #970, opened 2020-12-14
GCW-3427 Adds Process actions can link to a source

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3427

### What does this PR do?
Add source to Process actions


shreyas098 (approved):

app/models/processing_destinations_lookup.js
```diff
@@ -0,0 +1,13 @@
...
+
+export default Model.extend({
+  name: attr("string"),
+  lookups: function() {
+    const allRecords = this.store.peekAll("processing_destinations_lookup");
+    return allRecords.map(record => ({
+      id: record.get("id"),
+      name: record.get("name")
+    }));
+  }
```
patrixr, line 12: Don't think that `lookups` method has its place in the model (also did a search, doesn't seem to be used?)

app/routes/items/detail.js
```diff
@@ -40,6 +40,9 @@ export default AuthorizeRoute.extend({
     }
     await this.store.findAll("restriction", { reload: true });
     await this.store.findAll("donor_condition", { reload: true });
+    await this.store.findAll("processing_destinations_lookup", {
+      reload: true
+    });
```
patrixr, line 45: All 3 of those here probably don't need to be reloaded every time we open item details.
Either:
- We preload them
- We load them when we open the items details page, but only the first time (cache them essentially)

app/mixins/item_actions.js
```diff
@@ -186,6 +197,10 @@ export default Ember.Mixin.create(AsyncMixin, {
       }
     },
 
+    onProcessingDestinationChange(destination) {
```
patrixr, line 200: `setProcessingDestination` makes more sense as a name

`onXXX` would mean that it has already happened, but this method actually does the set

patrixr (changes requested):
Looks good, few minor comments

app/models/processing_destinations_lookup.js
bharat619: yeah. I had added it as some experimentation. Then forgot to remove. Thanks, removed it

app/routes/items/detail.js
bharat619: cached it

patrixr (approved):

PR #1012, opened 2021-04-16
Gcw 1957 Adding timer for resend pin

shreyas098:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3576

### What does this PR do?

FEATURE: Adding timer to send Pin


swatijadhav (approved):

app/controllers/authenticate.js
```diff
@@ -11,12 +11,28 @@ export default GoodcityController.extend(preloadDataMixin, {
...
+      this.set("pinAlreadySent", false);
+      this.set("otpResendTime", config.APP.OTP_RESEND_TIME);
+      return false;
+    }
+    this.set("timer", waitTime);
+    this.set("otpResendTime", waitTime - 1);
+    return setTimeout(() => {
+      this.timerFunction();
+    }, 1000);
+  },
```
patrixr, line 34: This looks alright, however I'm not sure why you need both `timer` and `otpResendTime` (can probably make do with a single one)

Aslo you should probably cancel the timeout when you navigate away from the page

patrixr (approved):

PR #1043, opened 2021-09-02
GCW-3700 preferences section with printer selection control

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3700

### What does this PR do?
- Account menu to set printer preferences and update details
- Added user-image component and used it in account-page and manage-users-page
- Moved common code to user-service

Screenshot 2021-09-02 at 6 24 10 PM


app/controllers/manage_account.js
```diff
@@ -0,0 +1,125 @@
...
+        this.get("user").rollbackAttributes();
+        Ember.$(`#${e.target.id}`).focus();
+
+        e.target.value
+          ? this.set(`${e.target.id}ValidationError`, true)
+          : this.set(`${e.target.id}InputError`, true);
+      }
+    },
+
+    changeAdminPrinter(value) {
```
shreyas098, line 101: Can we have a common function like
UpdatePrinter(event){
//since API call is same and we can pass tag as a param or make use of _event_ value and get that
}

swatijadhav: It is already a common API call added in `printer-service` to update printer based on tag of the application.

shreyas098 (approved):

PR #1059, opened 2021-10-08
Azure storage deployment

patrixr:
## Changes

- CI uses Azure cli to push the build to the goodcity storage account
- Web release now part of the Fastfile, command `STAGE=staging bundle exec fastlane web deploy`
- Release utils added (imported from Goodchat) to be used in Fastfile

@steveyken Probably better to "Squash and merge" this one :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@
 
 # fastlane
 fastlane/*.json
+fastlane/.byebug_history
 
 # misc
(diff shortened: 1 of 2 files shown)


patrixr:
> Awesome! Thank you. Not sure we need fastlane/.byebug_history but otherwise, good to go!

Already removed it : )

PR #1095, opened 2025-03-04
Cordova update

stanleylhs:
Upgrade Cordova Android to 13

Upgraded Plugin
Scandit to 7.1

Removed Photoviewer and its action


steveyken (approved):
Thank you!

